PR #52, opened 2017-03-25
Use getfield instead of eval for Packed type conversion

tkelman:
eliminates the need to overwrite these methods in other modules

as discussed at https://github.com/dehann/RoME.jl/commit/e564939b4db2bc02d26d5d9b21e2c338ec6815f4#commitcomment-21468647


dehann:
Ah great thanks!

dehann:
fixes issue on JuliaLang/METADATA.jl#8448

PR #115, opened 2018-07-12
general maintenance and new eval feature

dehann:
work in progress

### DIFF
--- a/src/FGOSUtils.jl
+++ b/src/FGOSUtils.jl
@@ -161,6 +161,20 @@ end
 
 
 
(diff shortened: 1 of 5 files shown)


src/FGOSUtils.jl
```diff
@@ -161,6 +161,20 @@ end
 
 
 
+function evalLikelihood(fg::FactorGraph, sym::Symbol, point::Vector{Float64})
```
dehann: relates to JuliaRobotics/Caesar.jl#136

PR #127, opened 2018-08-17
Feature/multithreadingminimize

dehann:


dehann:
In some cases `FunctorPairwiseMinimize` was shuffling order unnecessarily.  This PR is in preparation for the multithreading upgrade.

PR #100, opened 2018-05-27
A small threading example

dehann:
Not ready to merge yet, but these upgrades should be merged soon.

PR #128, opened 2018-08-20
multithreaded convolutions are working

dehann:
Code can still be cleaned up in many ways, but this is PR is the first major conversion to multiprocess / mutlithreaded approximate convolutions and inference in IIF.  The code relies solely on new `CommonConvWrapper` fucntors.  This PR is intended strictly for `IIF v0.3.7+` and beyond.

fyi @mc2922, @GearsAD

### DIFF
--- a/src/ApproxConv.jl
+++ b/src/ApproxConv.jl
@@ -1,4 +1,6 @@
 
+
+
 """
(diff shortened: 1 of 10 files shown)


dehann:
uses content and replaces #100

PR #131, opened 2018-08-27
Renormalize categorical probabilities when parsing multi hypothesis factors

keevindoherty:
Fixes issue in categorical distribution parsing by normalizing categorical probabilities.


dehann:
Thanks!

Issue #134, opened 2018-08-27
Node label conventions


pvazteixeira:
Currently, node labels are [assumed](https://github.com/JuliaRobotics/IncrementalInference.jl/blob/a44a082c497bea0c7a5ff5ee7fc8d6264c1af78c/src/FGOSUtils.jl#L82) to start with either `x` or `l` labels, followed by numbers only. It would be useful to have a more flexible convention on node labels, supporting, for instance, things like `p74_31`

pvazteixeira:
Changes in #139 help preserve ordering when only numerical indexing is used.

pvazteixeira:
#141 is great! I think for now it's OK to state the conventions/assumptions in node labeling (`l`,`x`, and indexing)

PR #136, opened 2018-08-28
Feature/nodenaming

pvazteixeira:
Addressing [#134](https://github.com/JuliaRobotics/IncrementalInference.jl/issues/134). `Pkg.test("IncrementalInference")` succeeds, but perhaps we should run more tests?


dehann:
Great, thanks!

PR #141, opened 2018-08-31
better sorting of ls for nested labels

dehann:


dehann:
improves #134 `ls` sorting

PR #156, opened 2018-09-07
oops fastnorm fix

dehann:


dehann:
see JuliaLang/julia#29091

Issue #148, opened 2018-09-04
Option to suppress all printouts


dehann:
In case user wants to see only their output -- won't be the default.

cc @mc2922

pvazteixeira:
#160 changes most of the `print` and `println` to `info`, which can be suppressed with `logging(DevNull; kind=:info)`. 

Unfortunately, the base logging facility is a bit coarse (supports `info`, `warn`, and `error`), so it would make sense to use something like [MicroLogging.jl](https://github.com/c42f/MicroLogging.jl)

dehann:
Also see new developments with CSM:
- https://github.com/JuliaRobotics/IncrementalInference.jl/blob/fcb17a53a885b62ab03a1da1d79a4147bdb97527/src/CliqStateMachine.jl#L18
- https://github.com/JuliaRobotics/IncrementalInference.jl/blob/fcb17a53a885b62ab03a1da1d79a4147bdb97527/src/SolveTree01.jl#L1592

PR #160, opened 2018-09-10
ability to supress most printouts

pvazteixeira:
almost `s:/print/info/g` and `s:/println/info/g`


dehann:
xref #148

dehann:
To disable logging you just need to add
```julia
logging(DevNull; kind=:info)
```
before using IncrementalInference functionalities.This will still let errors and warnings through. Please feel free to add to the comment (I wouldn't close it yet).

dehann:
also influences #224

PR #184, opened 2018-11-06
fast-forward ccolamd PR

dehann:


dehann:
will likely close, rather than merge this PR

PR #190, opened 2018-11-19
add TYPEDEF everywhere, change to addVariable!

dehann:


dehann:
also change addNode! to addVariable!

GearsAD:
:+1: for addVariable namechange!

PR #216, opened 2019-02-13
add warn for var duplication

### LINKED ISSUE #188: fail when trying to add variable that already exists in factor graph
Current code results in a bad graph with weird errors.  Could also just add a test rather than hard fail if user is unsure.

dehann:
fixes #188


dehann:
cc @GearsAD @Affie

dehann:
Fixes JuliaRobotics/Caesar.jl#252

PR #223, opened 2019-02-19
partial fix #222 and convenience features

### LINKED ISSUE #222: Check ordering of iterval IDs during `fmcmc!`
Possible that more optimal orderings are possible

dehann:


dehann:
partial fix to #222 

cc @GearsAD , @Affie

PR #225, opened 2019-02-20
general maintenance and new convenience functions

dehann:


dehann:
helps resolve JuliaRobotics/Caesar.jl#264

Issue #229, opened 2019-03-04
some use-cases want autoinit to be a separate process


dehann:
allows fast construction of graph skeleton

dehann:
add as init schema in clique solve order.

dehann:
will help fix #135 from user experience standpoint

dehann:
will help fix #124

dehann:
based on work from branch `feature/treeinitorder`.  PR to appear soon.

dehann:
significant changes to how solver handles these cases: therefore moved to new minor version v0.6.0

PR #298, opened 2019-05-29
Cliqsm/multiprocup

dehann:


dehann:
final parity with previous tree solution, fixes issues like:
- #229

PR #309, opened 2019-06-09
Partial fix testing, JLD2 bypassed for now

dehann:
IIF tests are passing with most of the code in tact.


dehann:
also dropping Julia 0.7 support

PR #328, opened 2019-06-24
Removing bayesnetvertid as it doesn't seem to be used

GearsAD:
Maybe leave this v0.8?


dehann:
This is fine for v0.7.x, but if we remove `BayesNetVertID` from the struct then we punt that PR to IIF v0.8.0

dehann:
see JuliaRobotics/DistributedFactorGraphs.jl#16

Issue #335, opened 2019-06-28
Cannot do Tree autoinit with clique containing only partial prior


dehann:
Need to request down init message instead.

PR #337, opened 2019-06-30
default autoinit=false

dehann:


dehann:
promotes solution on #335

Issue #339, opened 2019-07-03
Indefinite pause on CSM


dehann:
Seems to occur just before downsolve in some cases.  Stochastic behavior.

cc @mc2922

PR #342, opened 2019-07-08
starting migration fncs to DFG

dehann:
Needs JuliaRobotics/DistributedFactorGraphs.jl#57


dehann:
requires DFG v0.2.1

dehann:
implements #344

dehann:
partially resolves #339 , but new "locks" are required in clique data for csm to work reliably -- will be done as separate PR

PR #346, opened 2019-07-10
use locks to avoid csm race conditions

dehann:


dehann:
intends to fix #339

dehann:
cc @mc2922

dehann:
didnt quite get to use locks in the way I wanted to and had to fix one of the errors with a timeout wait instead

Issue #354, opened 2019-07-19
Collection of (fresh code) CSM related init/sove issues 


dehann:
- [ ] https://github.com/JuliaRobotics/RoME.jl/blob/examples/19Q3/beehive/examples/Beehive2D_ParialInitIssue.jl
- [ ] https://github.com/JuliaRobotics/Caesar.jl/blob/dynSAS/examples/marine/asv/kayaks/testDynPtExFail.jl
- [ ] https://github.com/JuliaRobotics/Caesar.jl/blob/dynSAS/examples/marine/asv/kayaks/rangeOnlySLAM_dbg.jl
- [ ] https://github.com/JuliaRobotics/Caesar.jl/blob/example/dfgsandshark/examples/marine/auv/Sandshark/Sandshark_AutoinitFailCase.jl

dehann:
These all are likely influenced by JuliaRobotics/DistributedFactorGraphs.jl#95

PR #356, opened 2019-07-22
Csm/3 q19/inferdim

dehann:
part solution to incorrect initialization during upsolve.  Still need aka "deep multi-branch sharing".

Partially address issue #354


dehann:
fixes #339

PR #357, opened 2019-07-24
parent sharing for down init

dehann:


dehann:
related to #354 example 4

dehann:
still debugging in all current test cases.

PR #358, opened 2019-07-31
test for hasfactor in parent sharing fg

dehann:


dehann:
fixes #370 but needs to lift limititers above 100 default

PR #360, opened 2019-08-02
add maxparallel to rmVarFromMarg

dehann:
cc @mc2922


dehann:
fixes oversight from #348

PR #366, opened 2019-08-04
prevent bss sampler from letting NaN through

dehann:


dehann:
bss now ignores SNRfloor if all weights are equal, so that it doesnt subtract to zero and try to divide by that zero.   Separate isnan test also introduced into bss constructor which errors on NaN

PR #367, opened 2019-08-05
Fixes for abstraction

GearsAD:


GearsAD:
FYI @Affie

Affie:
@dehann I would suggest that we take a smaller step like this and then do the cliqSubFg in a separate PR. DFG can provide the conversions quite easily for now.

dehann:
merge this first and then fix the likely conflicts with #369

Affie:
> merge this first and then fix the likely conflicts with #369

Currently, there is no conflicts. may change if this pr changes same files as #369

PR #369, opened 2019-08-07
LightGraphs(MetaGraphs) option as cliq sub fg

Affie:
Separate PR before I break everything


src/JunctionTreeTypes.jl
```diff
@@ -3,6 +3,8 @@
 
 const BTGdict = GenericIncidenceList{ExVertex,Edge{ExVertex},Array{ExVertex,1},Array{Array{Edge{ExVertex},1},1}}
 
+#supported in memmory fg types
+const InMemmoryDFGTypes = Union{DFG.GraphsDFG,DFG.LightGraphsDFG}
```
dehann: Memmory -> Memory

src/JunctionTreeTypes.jl
```diff
@@ -219,7 +221,7 @@ end
 $(TYPEDEF)
 """
 mutable struct FullExploreTreeType{T, T2}
-  fg::DistributedFactorGraphs.GraphsDFG
+  fg::InMemmoryDFGTypes
```
dehann: probably should template this one too

src/SubGraphFunctions.jl
```diff
@@ -53,10 +53,12 @@ Related
...
 getVariableIds
 """
-function buildSubgraphFromLabels(dfg::G, syms::Vector{Symbol}) where G <: AbstractDFG
+function buildSubgraphFromLabels(dfg::G, 
+                                 syms::Vector{Symbol}, 
+                                 destType::Type{<:AbstractDFG}=LightGraphsDFG ) where G <: AbstractDFG
 
   # data structure for cliq sub graph
-  cliqSubFg = initfg()
+  cliqSubFg = initfg(destType)
```
dehann: I made a small change here, hope that is okay?

dehann: work on copy for now and yes keeping same type is probably better, you're right.

PR #372, opened 2019-08-14
drop dfg requirement to 0.2.4

dehann:


dehann:
hard dependency on LightGraphsDFG

PR #380, opened 2019-09-09
LightGraphDFG was renamed to MetaGraphDFG

GearsAD:
Only merge this when DFG 0.4 is tagged.


dehann:
requires JuliaRobotics/DistributedFactorGraphs.jl#118

dehann:
resolves JuliaRobotics/RoME.jl#190

Issue #384, opened 2019-09-10
Downsolve, check for structure between multiple frontals


dehann:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/22c22bd65502c66c6dc9f5b0d4a5a8a9eee51ac9/src/SolveTree01.jl#L805

follows from #244

Issue #162, opened 2018-09-15
Weird results on bearing range factor (aka Funky21)


pvazteixeira:
Adding a range+bearing factor on `:x21` results in a strange solution, in which `:x25` moves away from its true value and there is a significant increase in uncertainty.

`drawPoseLandms`:
![drawposeslandms](https://user-images.githubusercontent.com/3402399/45582047-308d8480-b877-11e8-90bc-09d356a271b9.png)
![drawposeslandms2](https://user-images.githubusercontent.com/3402399/45582048-308d8480-b877-11e8-9a2e-b810b3a7b59a.png)
![drawposeslandms3](https://user-images.githubusercontent.com/3402399/45582049-308d8480-b877-11e8-91ef-2be1fc76d3e1.png)

Tree:
![btbug-ss-21](https://user-images.githubusercontent.com/3402399/45582050-35523880-b877-11e8-8349-beaa12c2f191.png)

The behavior is reproducible.

dehann:
might be related to #57

dehann:
repeat note:  The issue is reproducible using this script:
https://github.com/JuliaRobotics/Caesar.jl/blob/41323d5defa03418b7b68557de0cbbbb132c339d/examples/marine/auv/Sandshark/Funky21Problem.jl#L179

dehann:
move to v0.4.5 to rapidly resolve NLsolve v3.0.0 upstream breaking changes.

dehann:
NLsolve v3.0.0 restored with #196

dehann:
This test behavior might change in versions beyond v0.5.0, especially by v0.5.5, owing to other changes made.  Important to still test against old code to understand where this error came from.

dehann:
xref #244

dehann:
Work is progressing in new CSM tree init code.  Targeting v0.7.4 or v0.7.5 to get all the required upstream and downstream changes made.

PR #388, opened 2019-09-13
down solve with frontal neighbors,

dehann:
and many general fixes all round


dehann:
resolves #162

dehann:
fixes #384

PR #411, opened 2019-10-12
add solveFactorMeasurements

dehann:


dehann:
hey @pvazteixeira , you might like this feature.

Issue #214, opened 2019-02-12
getBestPointVal(fg, :x13), when to use max, mean, or on manifold moment


dehann:
Should simplify down to a single function that returns the best point value estimate for some variable.

dehann:
Considering both (but must work on-manifold):
* https://github.com/dehann/IteratedProbabilisticClassification.jl
* https://github.com/davidavdav/GaussianMixtures.jl

dehann:
see `calcVariablePPE`, and `MeanMaxPPE`

PR #421, opened 2019-10-22
Feature/update variable estimates

Affie:


src/FactorGraph01.jl
```diff
@@ -737,6 +737,8 @@ function doautoinit!(dfg::T,
         end
         pts,inferdim = predictbelief(dfg, vsym, useinitfct, logger=logger)
         setValKDE!(xi, pts, true, inferdim)
+        #TODO test
+        setVariablePosteriorEstimates!(xi)
```
Affie: @dehann, I think adding extra data breaks compare test, but I haven't had time to dig into the details.

dehann: these compare functions recently moved to DFG, so we should just tie up there too -- thanks!

dehann:
Hi @GearsAD , just mention you to keep mental note prior to DFG v0.5.0 -- that we may need to modify some of the compare stuff either in IIF or DFG when adding this PR.  I think this PR should land in the IIF v0.8.0 release, unless there are any objections.

Affie:
Lets hold off on this one and do it all in one go pending https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/147#issuecomment-542091523 I think we are close enough to a final decision.
I use it currently, but I'm happy working on a branch for now.

Affie:
needs https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/183

dehann:
addresses #214

Issue #427, opened 2019-11-11
CSM runaway iter in certain multihypo situations


dehann:
During `solveTree!` certain situations seem to result in infinite cycling of CSM:
- Specific case seems to be sibling cliques waiting on each other to progress with init/upsolve.

Related:
- related to JuliaRobotics/RoME.jl#199

dehann:
One case occurs as a race-condition when selecting which variable to initialize from an n-ary factor connected to several uninitialized variables.

dehann:
Multihypo case example, only `l1` and `l2` are initialized, problem is that a one by one initialization of other four variables cannot be achieved directly.

![Screenshot from 2019-11-11 08-40-15](https://user-images.githubusercontent.com/6412556/68591564-ff025000-045e-11ea-94b2-59f704a704cd.png)

Suggestions are:
- In multihypo case, allow the factor to be used in single-hypothesis mode for initialization, and then revert to defined multi-hypo,
- ...

Issue #248, opened 2019-04-02
Generic N-ary multihypothesis (bad-init nullhypo addition)  


dehann:
Binary hypothesis factors:
```julia
addFactor!(fg, .., multihypo=[1.0; 0.5; 0.5])
```
Produce two options where one must be true, and this is known to work well with sequential Gibbs sampling.

When more options are added, such as:
```julia
addFactor!(fg, .., multihypo=[1.0; 1/3;1/3;1/3])
```
sequential Gibbs sampling (with automated conditional construction) can arrive at inconsistent proposal distributions.  The clearest example is from 3-door identified in #236:
- three priored landmarks `(:l0,:l1,:l2)` at 0, 10, 40.
- a pose `x0` at 0.
- single variable sampling where trying to predict `:l1` or `:l2` from other information
```math
1/P  [L1 | X0, L0, L2, Z]
```

Question is if it impossible to find the conditional `[ | ]` without the partition function `1/P` that is guaranteed to always produce consistent proposals?

dehann:
Update:  i think the desired behavior is compromise between the legacy nested Gibbs strategy and introduction of two nullhypothesis strategies.  

Assumed solution at this point is expand the generalized multihypo API to automatically include an internal hypothesis---i.e. N+1 options---which represents stale particles from the current variable.  This avoids programming absolute (or even relative) data about uncertain variables into the factor.  This approach does change the proposal conditionals in Gibbs strategy to depend on old values of a proposed variable, but is limited to only one of (N+1) options.

If `multihypo=[1.0;0.5;0.5]`, the internal code should increase to `[1.0, 1/3,1/3,1/3]` where the last option is treated as keep stale, null hypothesis option.

dehann:
The rule of thumb is to multiply the uncertain values with the additional hypothesis -- i.e. binary becomes `2/3 * [p1, p2]`, or trinary becomes `3/4*[p1,p2,p3]`, etc.  The last 'new' null hypothesis mode will for the time being be set to `1/(N+1)` likelihood.

dehann:
fyi @keevindoherty

dehann:
required for JuliaRobotics/Caesar.jl#141

dehann:
moving this back (hopefully the last time) until after the DFG and CSM upgrades are sufficiently complete in IIF v0.7.0.  Several (maybe slightly messy) cleanups to occur throughout IIF v0.7.x.

dehann:
[...]

PR #429, opened 2019-11-11
fix 427 init with multihypo

dehann:


dehann:
fixing #427

dehann:
will influence design for #248

PR #434, opened 2019-11-18
Fix for tests - note that having different estimates will fail in compare

GearsAD:


test/testCompareVariablesFactors.jl
```diff
@@ -52,15 +53,17 @@ x1b = getVariable(fg2, :x0)
 
 ensureAllInitialized!(fg2)
 
-@test compareSimilarVariables(fg, fg2, skipsamples=true, skip=Symbol[:initialized;:inferdim])
+@test compareSimilarVariables(fg, fg2, skipsamples=true, skip=Symbol[:initialized;:inferdim;:estimateDict])
+# fg2 has been solved, so it should fail on the estimate dictionary
+@test !compareSimilarVariables(fg, fg2, skipsamples=true, skip=Symbol[:initialized;:inferdim])
```
dehann: could consider adding a separate test or jack up estimateDict testing for general case in the future.

PR #439, opened 2019-11-18
Update ccolamd.jl

dehann:


dehann:
an example
```julia
using SparseArrays
using SuiteSparse.CHOLMOD: SuiteSparse_long

As = SparseMatrixCSC(rand(10,10) .< 0.2)
cons = zeros(SuiteSparse_long, length(As.colptr) - 1)
cons[end] = 10

Ccolamd.ccolamd(As, cons)
```

BUT did not immediately work, think we just missed something silly.

src/ccolamd.jl
```diff
@@ -76,16 +76,15 @@ function ccolamd!(n_row, A::Vector{SuiteSparse_long}, p::Vector{SuiteSparse_long
...
+    return ccolamd!(n_row, A, p, constraints)
 end
 
-ccolamd(n_row, A::Vector{SuiteSparse_long}, p::Vector{SuiteSparse_long}) =
-    ccolamd!(n_row, copy(A), copy(p))
+ccolamd(n_row, A::Vector{SuiteSparse_long}, p::Vector{SuiteSparse_long}, constraints=zeros(SuiteSparse_long, length(p) - 1)) =
+    ccolamd!(n_row, copy(A), copy(p), constraints)
 
-ccolamd(A::SparseMatrixCSC) = ccolamd(size(A, 1), A.rowval, A.colptr)
+ccolamd(A::SparseMatrixCSC, constraints=zeros(SuiteSparse_long, length(A.colptr) - 1)) = ccolamd(size(A, 1), A.rowval, A.colptr)
```
tonioteran: The `constraints` variable, although exposed, is not being used to call the `ccolamd` function. I think it should be:
```jl
... = ccolamd(size(A,1), A.rowval, A.colptr, constraints)
```
with the `constraints` variable at the end.

dehann: oh good catch thanks -- will update now

dehann:
i made the change thanks, but there is another issue -- looks like a pointer or vector sizing issue.  Will dig a little deeper to fix.

PR #442, opened 2019-11-20
parametric structures

Affie:


dehann:
this is starting to address JuliaRobotics/Caesar.jl#140

src/JunctionTreeTypes.jl
```diff
@@ -14,14 +14,17 @@ mutable struct BayesTree
   btid::Int
   cliques::Dict{Int,Graphs.ExVertex}
   frontals::Dict{Symbol,Int}
+  #TEMP JT for evaluation, to store message channels in tree, would be better stored on edges
```
dehann: thats fine, continue on that path.  I think the intermediate solution will be to use wrapper function `getCliqMsgUp` etc.

src/BeliefTypes.jl
```diff
@@ -41,3 +41,41 @@ mutable struct NBPMessage <: Singleton
...
+# mutable struct NBPMessage <: Singleton
+#   status::Symbol # Ek kort die in die boodskap
+#   p::Dict{Symbol, EasyMessage}
+# end
+
+struct TreeBelief
+  val::Array{Float64,2}
+  bw::Array{Float64,2}
+  inferdim::Float64
+  # manifolds::T TODO ? JT not needed as you have the variable with all the info in it?
```
dehann: yeah, agree that we have the manifolds from softtypes in the variables themselves, which should be plenty to get the ball rolling.

src/FactorGraphTypes.jl
```diff
@@ -58,6 +58,7 @@ mutable struct SolverParams <: DFG.AbstractParams
   N::Int
   multiproc::Bool
   logpath::String
+  algorithms::Vector{Symbol} # list of algorithms to run [:default] is mmisam
```
dehann: think this is pretty cool and will be very useful for prototyping and dev work too -- great idea thanks!

src/ParametricCliqStateMachine.jl
```diff
@@ -0,0 +1,337 @@
+# STATUS messages [:initialized;:upsolved;:marginalized;:downsolved;:uprecycled]
```
dehann: sorry, yeah we need to document the CSM better.  Guess we can start here JuliaRobotics/Caesar.jl#440

src/ParametricCliqStateMachine.jl
```diff
@@ -0,0 +1,337 @@
...
+- Parametric state machine function nr. 3
+"""
+function solveUp_ParametricStateMachine(csmc::CliqStateMachineContainer)
+
+  infocsm(csmc, "Par-3, Solving Up")
+
+  setCliqDrawColor(csmc.cliq, "red")
+  csmc.drawtree ? drawTree(csmc.tree, show=false, filepath=joinpath(getSolverParams(csmc.dfg).logpath,"bt.pdf")) : nothing
+
+  # is dit nodig om vir parametric die factors by te sit?
```
dehann: in the current mmisam solution the factors symbols are recorded in the BayesTreeNodeData for each clique, but actual factors are only fetched from the distributed factor graph object when the cliqSubFg is created.  Perhaps okay to follow that pattern here too for the time being...

src/ParametricCliqStateMachine.jl
```diff
@@ -0,0 +1,337 @@
...
+  # store the cliqSubFg for later debugging
+  opts = getSolverParams(csmc.dfg)
+  if opts.dbg
+    DFG.saveDFG(csmc.cliqSubFg, joinpath(opts.logpath,"logs/cliq$(csmc.cliq.index)/fg_beforeupsolve"))
+    drawGraph(csmc.cliqSubFg, show=false, filepath=joinpath(opts.logpath,"logs/cliq$(csmc.cliq.index)/fg_beforeupsolve.pdf"))
+  end
+
+  #TODO UpSolve cliqSubFg here
+  # slaap om somme te simileer
+  sleep(rand()*2)
```
dehann: this is where we call the JuliaNLSolvers routines to do the actual matrix computations on cliqSubFg -- or better yet just a call to batchSolveParametric!(csmc.cliqSubFg) as you had suggested before.

src/ParametricCliqStateMachine.jl
```diff
@@ -0,0 +1,337 @@
...
+  @info "$(csmc.cliq.index): Solve Finished"
+
+  if isa(csmc.dfg, DFG.InMemoryDFGTypes)
+    #TODO update frontal variables here directly
+
+    #solve finished change color
+    setCliqDrawColor(csmc.cliq, "lightblue")
+    csmc.drawtree ? drawTree(csmc.tree, show=false, filepath=joinpath(getSolverParams(csmc.dfg).logpath,"bt.pdf")) : nothing
+
+    @info "$(csmc.cliq.index): Finish en klaar"
```
dehann: lol

dehann:
@Affie , comments might not have made it through on Github because it looks like i have to submit the comments as review first.  Sorry if you had been waiting on previous comments.  I just added a few here now.  This is great thanks!

src/BeliefTypes.jl
```diff
@@ -51,3 +51,6 @@ struct ParametricBelieveMessage <: AbstractBeliefMessage
   belief::Dict{Symbol, Vector{Float64}}
   inferdim::Dict{Symbol, Float64} #dink dis die ene wat Dehann soek
 end
+
+ParametricBelieveMessage(status::Symbol) =
```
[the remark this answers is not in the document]
Affie: fixed

src/ParametricCliqStateMachine.jl
```diff
@@ -91,9 +93,31 @@ function waitForUp_ParametricStateMachine(csmc::CliqStateMachineContainer)
...
+  inlist = Edge{V}[]
+  for edgelist in inclist
+    for ed in edgelist
+      if ed.target.index == targid
+        push!(inlist, ed)
+      end
+    end
+  end
+  return inlist
+end
```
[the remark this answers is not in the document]
Affie: ok will do

src/FactorGraphTypes.jl
Affie: still a work in progress, but it seems to work

src/ParametricCliqStateMachine.jl
Affie: Good idea, that will help. If this is more defined we can maybe use enumerated types?

src/ParametricCliqStateMachine.jl
Affie: I think it will work to start with. I like the idea of `batchSolveParametric!(csmc.cliqSubFg)` since you can also use it to solve bigger factor graphs  and then do the incremental in the tree.

Affie: We can also look at computing the conditionals

src/ParametricCliqStateMachine.jl
```diff
@@ -91,9 +93,31 @@ function waitForUp_ParametricStateMachine(csmc::CliqStateMachineContainer)
 
   infocsm(csmc, "Par-2, wait for up messages of needed")
 
+  for e in Graphs.out_edges(csmc.cliq, csmc.tree.bt)
+    @info "$(csmc.cliq.index): take! on edge $(e.index)"
+    # Blocks until data is available.
+    believeMsg = take!(csmc.tree.messages[e.index].upMsg)
+    @info "$(csmc.cliq.index): Believe message recieved with status $(believeMsg.status)"
+    #TODO save up message and add priors to cliqSubFg
+  end
```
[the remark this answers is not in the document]
Affie: will do

src/ParametricCliqStateMachine.jl
Affie: Good to know, I think that is why I couldn't remove the duplicate to test if it is a bug.

src/ParametricCliqStateMachine.jl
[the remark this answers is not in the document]
Affie: Will do with other functions. For now I like having it close. But I know it makes it harder for someone to follow.

src/ParametricCliqStateMachine.jl
dehann: Symbols are like enumerated types (C++) or are you talking about hard types -- just want to avoid having too many abstract types floating around?

src/ParametricCliqStateMachine.jl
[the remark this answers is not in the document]
dehann: however and whenever convenient, no rush

src/ParametricCliqStateMachine.jl
dehann: Computing the conditionals will definitely be of interest, actually expect @tonioteran might be interested in that.  There is another name in the lingo called "smart factors" (L. Carlone et al.)

src/ParametricCliqStateMachine.jl
Affie: Neither, https://docs.julialang.org/en/v1/base/base/#Base.Enums.@enum

Affie: I would prefer `enum` over `Symbol` for ease of use and docstrings
```julia
julia> IIF.CliqStatus
Enum IncrementalInference.CliqStatus:
initialized = 0
upsolved = 1
marginalized = 2
downsolved = 3
uprecycled = 4

help?> IIF.CliqStatus
  primitive type CliqStatus <: Enum{Int32} 32

  Clique status message enumerated type
```

src/BeliefTypes.jl
```diff
@@ -66,16 +66,24 @@ end
 TreeBelief(p::BallTreeDensity, inferdim::Real=0.0) = TreeBelief(getPoints(p), getBW(p), inferdim)
 TreeBelief(val::Array{Float64,2}, bw::Array{Float64,2}, inferdim::Real=0.0) = TreeBelief(val, bw, inferdim)
 
+"""
+    CliqStatus
+Clique status message enumerated type with status:  
+initialized, upsolved, marginalized, downsolved, uprecycled
+"""
+@enum CliqStatus initialized upsolved marginalized downsolved uprecycled
```
dehann: oh cool, didn't know you can do that.  There is a small chance that these are Symbol under the hood, but either way this is a much cleaner representation.

Affie: It defaults to Int32 see comment : https://github.com/JuliaRobotics/IncrementalInference.jl/pull/442#discussion_r353575325

Issue #438, opened 2019-11-18
[FEATURE] special real-time odo factor and variable (adjacent to SLAM solve)


dehann:
![IMG_2277](https://user-images.githubusercontent.com/6412556/69084312-f9ce7380-0a11-11ea-9e3a-c9ad6daa2252.jpg)

dehann:
additional feature is to snap rt-odo to regular odo once integration between consecutive poses has completed -- i.e update fg object at high rate, regardless of other SLAM processes on graph.

dehann:
should just standarize this as pseudo ODE solver, which also handles things like continuous to discrete process noise work (Qc -> Qd) etc.

dehann:
needs JuliaRobotics/DistributedFactorGraphs.jl#201

PR #448, opened 2019-11-25
fixing solvable

dehann:


dehann:
part of JuliaRobotics/Caesar.jl#433

dehann:
part of #438

dehann:
see #450 instead

Issue #458, opened 2019-11-27
Confirm clique sub-graphs only use priors once 


Affie:
### The tree
![image](https://user-images.githubusercontent.com/6612981/69703737-c7e89b80-10fa-11ea-8914-96dd1b4083ea.png)

### fg_built, see `l0f1`
![image](https://user-images.githubusercontent.com/6612981/69703347-db473700-10f9-11ea-9e6e-530d73c7ff0b.png)

### fg_beforeupsolve look like it is used
![image](https://user-images.githubusercontent.com/6612981/69703329-d1bdcf00-10f9-11ea-8633-50e96c2f3650.png)

dehann:
yes this is a bug, the `:l0f1` should only occur once.  have to double check if it should first occurrence or frontals only.  The hard part is how do this for initialization.

Affie:
By the same logic: Should all factors not only be used once? On frontals.
see `x3lm3f1`
Note, I'm only looking at subgraphs fg_beforeupsolve (message priors included) so don't know what else the code might do.
**Tree**
![image](https://user-images.githubusercontent.com/6612981/70305478-2f929b00-180d-11ea-9c0f-2d6d1104a1ff.png)
**Clique 1 - x3,lm0,lm3,x1**
![image](https://user-images.githubusercontent.com/6612981/70305258-a5e2cd80-180c-11ea-99c7-662d6bd73852.png)

**Clique 2 - x2: x1,x3,lm3**
![image](https://user-images.githubusercontent.com/6612981/70305435-0c67eb80-180d-11ea-99e4-958582faaf95.png)

Affie:
just copying here for now:
```julia
# create subgraph with no factors on/between seperators
function buildSubgraphFromLabels!(dfg::AbstractDFG,
                                  cliqSubFg::AbstractDFG,
                                  frontals::Vector{Symbol},
                                  separators::Vector{Symbol};
                                  solvable::Int=0)

  for sym in separators
    DFG.addVariable!(cliqSubFg, deepcopy(DFG.getVariable(dfg, sym)))
  end

  addfac = Symbol[]
  for sym in frontals
    DFG.addVariable!(cliqSubFg, deepcopy(DFG.getVariable(dfg, sym)))
    append!(addfac, getNeighbors(dfg,sym))
  end

  allvars = ls(cliqSubFg)
  for sym in addfac
    fac = DFG.getFactor(dfg, sym)
[...]

Issue #221, opened 2019-02-18
Possible strange behavior when mixing large and small noise factors


dehann:

dehann:
blocked behind DFG upgrade for the short term

PR #461, opened 2019-12-03
simple prior use test

Affie:
I feel this test should pass


dehann:
Hi, oh great thanks will resolve today!

dehann:
hi, didnt get through this today, but will look first thing in the morning.

test/priorusetest.jl
```diff
@@ -0,0 +1,83 @@
...
+        tree, smt, hist = solveTree!(fg)
+        x0_m = getKDEMean(getKDE(getVariable(fg, :x0)))[1]
+        x1_m = getKDEMean(getKDE(getVariable(fg, :x1)))[1]
+        x2_m = getKDEMean(getKDE(getVariable(fg, :x2)))[1]
+
+        @info ("Testing means = 0 with 2 priors:\ngraphinit=$graphinit\nMeans: x0: $(x0_m), x1: $x1_m, x2: $x2_m")
+
+        @test isapprox(x0_m, 0.0, atol = 0.1)
+        @test isapprox(x1_m, 0.0, atol = 0.1)
+        @test isapprox(x2_m, 0.0, atol = 0.1)
```
dehann: @Affie, first question: if you are putting two priors at the ends of a 3 variable chain `(x0,x1,x2)`, either at + or - 1 -- why would you expect them to all be at average 0?  I would expect them to be at + or 1 one respectively, with `x1` floating somewhere 0 in the middle?

This was based on me reading the code.

dehann:
## first test graph

Screenshot from 2019-12-04 20-07-58

Affie:
I don't agree on changes.
What about this: changing the variable ordering changes the answers.
image

GraphInit = true
Means: x0: 0.031117603087722025, x1: 0.1416983152637742, x2: 0.23945144191614984
Means: x0: -0.41846814644201763, x1: -0.3480670649963539, x2: -0.1642135339004435

GraphInit = false
Means: x0: 0.07968446871195635, x1: 0.14465511208606313, x2: 0.22910276233886342
Means: x0: -0.33884567700033713, x1: -0.281886695060186, x2: -0.19827585825173885

Affie:
If #221 is the issue I brought up it is related.
And second graph tests for #458

test/priorusetest.jl
dehann: x0 and x2 should be near 0, and only x1 should be "on" 0.  Test with loose `atol=0.1` bit hacky, but okay :-)  x0 and x2 should be around 0.1 or something like that.

dehann:
adding @tonioteran here for interest sake.

I strongly agree that changing the variable ordering should not change the numerical result.  Thanks for describing, I definitely need to spend time on this and make sure we understand what is happening here.

test/priorusetest.jl
Affie: Please note, I changed the standard deviation to 0.01.
So its more like -0.01, 0, 0.01 for x0,x1,x2.

-1/1 <-x0-> 1/0.01 <-x1-> 1/0.01 <-x2-> 1/1

Affie:
I looks like it either has to do with the structure or the priors on the down solve.

test/priorusetest.jl
dehann: i'm assuming you did the calculation for multiplying two Gaussians, and found that the new mean should be at +- 0.01?  What I mean is if you were to multiply N(-1,0.1) * N(+1,0.1), the answer will be at zero with a new covariance slightly smaller than 0.1.

dehann:
acknowledged thanks, definitely will do a deep dive on this soon.

dehann:
xref #462

dehann:
xref #485

dehann:
xref #487

Issue #467, opened 2019-12-05
Consolidate / Simplify factor evaluation (CalcFactor)


dehann:
EDITED (20Q4 / 21Q1): 

Using this issue for topics related to consolidating the nonparametric and parametric factor residual function API.  Idea is user writes one residual factor in a sensible way, and IIF is able to use that factor for both parametric and nonparametric with little to no change from the user perspective.  **A critical point to note** here is that nonparametric (nonGaussian) is much more versatile and general, so there are many many cases where parametric will just not be possible -- regardless the API should be standardized.

#### References:
- Similar for `IIF.getSample` #927
- Adjacent issue #1025 
- TODO docs #1044 
- Towards closing JuliaRobotics/RoME.jl#244
  - Keeping in mind: https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/issues/41#issuecomment-733636425

------

### Current holdups:
- [x] https://github.com/JuliaRobotics/IncrementalInference.jl/issues/467#issuecomment-753656505
- [x] Nonparametric `AbstractRelativeMinimize` should return `residual::Vector`, not self accumulate to a Real scalar value.

------
### New Documentation
- https://juliarobotics.org/Caesar.jl/latest/examples/basic_definingfactors/

------

### Preliminary New API Example

- also see https://github.com/JuliaRobotics/IncrementalInference.jl/issues/467#issuecomment-751867661

https://github.com/JuliaRobotics/IncrementalInference.jl/blob/0469062bc1599222233801d75c6b546e5ae10ea5/src/Factors/LinearRelative.jl#L42-L51

------

### Parametric Integration Work 90% (?) Done

DF, I was hoping to just finish the consolidation with (https://github.com/JuliaRobotics/IncrementalInference.jl/issues/467#issuecomment-753656505) but alas!  Note that much of the consolidation work has already been done surrounding `_CalcFactorParametric` as the internal interface between new `CalcFactor` API and `solve__Parametric` functions.  E.g. see:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/9e225b7a70763ea8df5a9f3b1bc4d3dcd328d88f/src/ParametricUtils.jl#L81-L90

------

### Original Issue Question

How would you use a normal IIF factor in parametric? example.

dehann:
cc @Affie

Affie:
Ok to implement cost directly
```julia
function (s::LinearConditional)(X1::Vector{Float64},
[...]

Issue #459, opened 2019-11-28
Consolidate belief messages and message channels 


Affie:
This is a summary:
- [x] 1. Messages
- [x] 2. Channels
  - [x] up (IIF v0.13)
  - [x] dwn (IIF v0.16, planned)

---
xref summary of related aspects:
- #385 (refactor tree messages up and dwn to single container type)
- #465 (VND.bw a 3 index array :heavy_multiplication_x: )
- #468 (first merge linear parametric solution)
- #558 (overview of issues relating to parametric and consolidation)
- #579 (generalized deconvolution)
- #600 (add covariance calculations to parametric)
- #647 (refactor LikelihoodMessage TreeBelief)
- #648 (drop TempBeliefMsg)
- #650 (fix test failure)
- #651 (further consolidate tree messages)
- #652 (fast-forward parametric dev)
- #653 (add Likelihoods rm BeliefMessage)
- #654 (refactor tree msg types)
- #673 (super solving)
- #674 (push pull model)
- #675 (store msg channels in csmc/btnd or separately in tree & edges)
- #855 (fetch vs. take!)
---

## 1. Likelihood messages
 - Message will be the marginalized likelihood of the separators in the clique
 - `variableOrder` are variable labels to identify the values
 - In the case of parametric an MvNormal can be used to store μ and Σ <: `SampleableBelief`
 - `inferdim`
 - `softtype` (for manifold and variable dimension)
 - Flag or other way to show the presence of prior information.
 - Earlier names include: productFactor, Fnew, MsgPrior, LikelihoodMessage

Types of factors made from messages so far
 - Prior factor (as existing)
 - Conditional factors constructed from the deconvolution of the likelihood message.
    - Need to identify the type such as `Pose2Pose2`, `Pose2Point2BearingRange`

#### Structure used for the Likelihood message WIP

```julia
"""
    CliqStatus
Clique status message enumerated type with status:
initialized, upsolved, marginalized, downsolved, uprecycled
"""
@enum CliqStatus INITIALIZED UPSOLVED MARGINALIZED DOWNSOLVED UPRECYCLED ERROR_STATUS
#we probably also need: CANT_INITIALIZE, PARTIAL_INIT, etc.
```
```julia
struct LikelihoodMessage{T <: SamplableBelief} #<: Singleton
  status::CliqStatus
  belief::Dict{Symbol, TreeBelief}
  variableOrder::Vector{Symbol}
[...]

PR #468, opened 2019-12-06
WIP Linear parametric solution

Affie:
Draft PR to track changes with master and discuss code
TODO
- Initialize #541
  - Additional states and messages will be required
- Solve 
  - Calculate covariances #539
  - Optim - linear works
  - NLSolve

basic usage
```julia
fg = LightDFG{SolverParams}( params=SolverParams(algorithms=[:default, :parametric]))
#add linear variables and factors ...
# Or generate a sample fg with lineStepFG
# fg = lineStepFG(7, poseEvery=1, landmarkEvery=3, posePriorsAt=collect(0:7), sightDistance=2)
#NOTE experimental and not exported. 
IIF = IncrementalInference
# bulk solve factor graph
d,st = IIF.solveFactorGraphParametric(fg)

# DEV NOTE
# For development on the tree solve (not currently working):
# build the tree
tree = wipeBuildNewTree!(fg)
# init the messages
IIF.initTreeMessageChannels!(tree)
# solve the tree
IIF.solveTreeParametric!(fg, tree)

```


src/BeliefTypes.jl
```diff
@@ -41,3 +41,49 @@ mutable struct NBPMessage <: Singleton
...
+# mutable struct NBPMessage <: Singleton
+#   status::Symbol # Ek kort die in die boodskap
+#   p::Dict{Symbol, EasyMessage}
+# end
+
+struct TreeBelief
+  val::Array{Float64,2}
+  bw::Array{Float64,2}
+  inferdim::Float64
+  # manifolds::T TODO ? JT not needed as you have the variable with all the info in it?
```
dehann: if `manifolds` was in the original struct then its probably definitely needed, please add.  We can do consolidation and reduction separately.  This field is probably used in a situation when the variable data is not readily available.

src/DefaultNodeTypes.jl
```diff
@@ -2,6 +2,9 @@
 
 SamplableBelief = Union{Distributions.Distribution, KernelDensityEstimate.BallTreeDensity, AliasingScalarSampler}
 
+#Supported types for parametric
+ParametricTypes = Union{Normal, MvNormal}
```
dehann: think we should only use `MvNormal` in all cases.

src/DefaultNodeTypes.jl
```diff
@@ -48,6 +51,32 @@ struct Prior{T} <: IncrementalInference.FunctorSingleton where T <: SamplableBel
 end
 getSample(s::Prior, N::Int=1) = (reshape(rand(s.Z,N),:,N), )
 
+
+# TODO maybe replace X with a type.
+#TODO ParametricTypes or dispatch on each? I like it together for now. Maybe compiler will split it?
+function (s::Prior{<:ParametricTypes})(X1::Vector{Float64};
+                    userdata::Union{Nothing,FactorMetadata}=nothing)
+
+  if isa(s.Z, Normal)
```
dehann: as above, rather use only MvNormal, rather then battle with API differences and `isa` vs multiple dispatch.

src/DefaultNodeTypes.jl
```diff
@@ -134,6 +186,32 @@ function (s::LinearConditional)(res::Array{Float64},
   nothing
 end
 
+# parametric specific functor
+function (s::LinearConditional{<:ParametricTypes})(X1::Vector{Float64},
```
dehann: I'm not fond of the parametric specific functors (the code maintainability cost will be way too high), rather do a wrapper on the existing so that functors are defined only once according to the existing standard:
https://www.juliarobotics.org/Caesar.jl/latest/concepts/adding_variables_factors/
https://www.juliarobotics.org/Caesar.jl/latest/examples/basic_definingfactors/

src/FactorGraph01.jl
```diff
@@ -351,6 +351,43 @@ function getOutNeighbors(dfg::T, vertSym::Symbol; needdata::Bool=false, ready::I
   return nodes
 end
 
+
+
+function DefaultNodeDataParametric(dodims::Int,
```
dehann: bread crumb for future consolidation with existing `setDefaultNodeData`

src/FactorGraphTypes.jl
```diff
@@ -58,6 +58,7 @@ mutable struct SolverParams <: DFG.AbstractParams
   N::Int
   multiproc::Bool
   logpath::String
+  algorithms::Vector{Symbol} # list of algorithms to run [:default] is mmisam
```
dehann: definitely worth giving it a try

src/IncrementalInference.jl
```diff
@@ -534,6 +534,11 @@ include("TetherUtils.jl")
 include("CliqStateMachine.jl")
 include("CliqStateMachineUtils.jl")
 
+#EXPERIMENTAL parametric
+include("ParametricSolveTree.jl")
+include("ParametricCliqStateMachine.jl")
+include("ParametricUtils.jl")
```
dehann: nice that its in three separate files, as we figure out how to consolidate and dispatch we can reduce the amount of code in these files down to a bear minimum through a few release cycles

src/JunctionTreeTypes.jl
```diff
@@ -14,14 +14,17 @@ mutable struct BayesTree
   btid::Int
   cliques::Dict{Int,Graphs.ExVertex}
   frontals::Dict{Symbol,Int}
+  #TEMP JT for evaluation, to store message channels in tree, would be better stored on edges
```
dehann: personally im not there yet for messages on edges, but havent worked with it enough to know how existing CSM workflows will have to change to accommodate both cliques and edges.  Previous workflow was to only use cliques and that became hairy enough.  The suggestion of "hiding" where the messages are stored through accessor API functions might work, and will have to watch out for debug and step processes in CSM dev where cliques are being cloned (but edges are currently not being cloned).

src/JunctionTreeTypes.jl
```diff
@@ -51,6 +54,8 @@ mutable struct CliqStateMachineContainer{BTND, T <: AbstractDFG, InMemG <: InMem
   refactoring::Dict{Symbol, String}
   oldcliqdata::BTND
   logger::SimpleLogger
+  parametricMsgsUp::Vector{ParametricBeliefMessage} #TODO TEMP net om te kyk hoe dit werk, dit moet eintelik <: abstact wees
+  parametricMsgsDown::Vector{ParametricBeliefMessage}
```
dehann: Again, I am quite worried about the <:Abstract concept being introduced everywhere.  There was the long discussion on anonymous static types, where the Julia guys were nervous.  We let it be `estimateDict` case, but I don't think we should be introducing it everywhere so quickly.  As with the comment above, I feel there should be a single BeliefMessage type that works for all cases -- although this will have to happen as part of tree message consolidation.

src/BeliefTypes.jl
Affie: Manifold is not in the current mm implementation. 
`TempBeliefMsg = Dict{Symbol, Tuple{BallTreeDensity, Float64}}`

src/DefaultNodeTypes.jl
Affie: I'll leave it as is for now to fit in with mm that currently works with `Normal` and `MvNormal`. We can change it on both at a later stage if you want.

src/DefaultNodeTypes.jl
Affie: Noted. Reasons for it was in slack, can't remember it now. 
I would suggest we keep it as is for now since it works. We can add tests and then swap over to ensure it works the same.
I'm slightly worried about the performance of the mm functors for parametric.

src/FactorGraph01.jl
Affie: That was the plan. For now I tried to make the code as detached as possible to ease merging at a later stage. I suggest we carry on in this way until it is merged. We can then do smaller branches to do consolidation.

src/JunctionTreeTypes.jl
Affie: Note, it is just the message channel associated with the edge (there is no more information added, the relationship already exists). The message data is stored in CSM.
So there is no need to clone edges.

src/JunctionTreeTypes.jl
Affie: <: abstract parametric, was to allow for different message types, I believe one message type will work.

dehann:
haha, sorry -- I removed the label based on PR being a 'Draft', but probably safer to have the label :-)

src/CanonicalGraphExamples.jl
```diff
@@ -104,3 +103,69 @@ function loadCanonicalFG_CaesarRing1D(;graphinit::Bool=false)
...
 end
+
+
+
+"""
+    $SIGNATURES
+Continuous, linear scalar and multivariate test graph generation. Follows a line
+with the pose id equal to the ground truth.
+"""
+function lineStepFG(lineLength::Int;
```
dehann: Want suggest perhaps staying with the naming convention of `loadCanonical_LIneStepFG` for now, if that is okay please?

Affie: No problem, Is `load` still correct as it generates a graph from parameters.

Affie:
> haha, sorry -- I removed the label based on PR being a 'Draft', but probably safer to have the label :-)

I was jut a bit worried that you wanted to merge it. It is keeping up to date with master so I think we can keep it separate for now. We should decide when to merge, I would like the tree to first work, but we need co-variances.

src/JunctionTreeTypes.jl
Affie: refer to #459 for further discussion.

src/BeliefTypes.jl
dehann: can't remember immediately but I do recall something about any code using `TempBeliefMsg` having to go back to a/the factor graph to `getManifolds` using the provided `Symbol` -- it was a schlep because it required a valid `dfg` object to be passed all the way down to that function...

src/DefaultNodeTypes.jl
dehann: okay

src/DefaultNodeTypes.jl
dehann: Right, so if there is a performance hit with the mm functions, that needs to be fixed immediately!  I spent a LOT of time on those functions to get good performance for where things were.  There are one or two memory things that can be better for the general `multihypo` interface, but that is a layer higher than what you see here, so should be fine.

i try get wrappers to you ASAP, because otherwise these things float far apart -- maybe little bit of slowly in the right direction type of argument.

dehann:
Swapping the tree, new CSM, and the inference in one big branch, sjoe okay!  You going to have to help with the consolidation the too please :-)

Affie:
> Swapping the tree, new CSM, and the inference in one big branch, sjoe okay! 
Jip, sorry. The tree foundation is in master, but I opened a can of worms with the clique index. So I first wanted to experiment and test a few ideas in the simpler parametric CSM.
I'm working towards reusing the tree. So we will need a `deleteClique` function.

We can merge sooner if you feel its needed. But we will end up with not fully functional code in a tag. This branch has not diverged that far from master yet, but its starting to get harder to keep it consolidated.

src/BeliefTypes.jl
Affie: Lets continue this discussion in #459

Affie: Its in struct, but not populated everywhere yet.

src/DefaultNodeTypes.jl
Affie: How would you feel about removing the `idx` parameter and only working with the vector (not the matrix). It makes it a bit easier and improves readability.

dehann:
targeting slightly premature merge just after IIF v0.8.4 is tagged, coming up soon.

src/DefaultNodeTypes.jl
Affie: #467

PR #472, opened 2019-12-11
sub-graphs with only factors on frontals connected

### LINKED ISSUE #458: Confirm clique sub-graphs only use priors once

Affie:
This may be a partial fix for the test and fixes #458 (possibly)

Do not merge without looking at the effect of not having factors in cliques


src/SolveTree01.jl
```diff
@@ -71,6 +71,7 @@ function packFromLocalPotentials!(dfg::AbstractDFG,
   #
   inferdim = 0.0
   for idfct in getData(cliq).potentials
+    !(exists(dfg, idfct)) && (@warn "$idfct not in clique $(cliq.index)" continue)
```
Affie: FIXME I just put this in without looking to get the code to run

src/SolveTree01.jl
```diff
@@ -96,6 +97,7 @@ function packFromLocalPartials!(fgl::G,
   #
 
   for idfct in getData(cliq).potentials
+    !(exists(fgl, idfct)) && (@warn "$idfct not in clique $(cliq.index)" continue)
```
Affie: FIXME I just put this in without looking to get the code to run

src/SubGraphFunctions.jl
```diff
@@ -87,6 +87,45 @@ function buildSubgraphFromLabels(dfg::G,
 
   return cliqSubFg
 end
+
+#TODO solvable
+function buildSubgraphFromLabels!(dfg::AbstractDFG,
```
Affie: PR is more about this function. Note its implemented and used in parametric

dehann:
hI @Affie , got it thanks!  In the default nonparametric solution we are (should) be using all factors connected to frontals during the downsolve, so those two should be consistent.  The current upsolve has one additional check to ensure that all the required variables for those factors are in the union of frontals and separators for that clique.

dehann:
In the downsolve case, all factors connected to the frontals implies the downsolve subgraph needs to have more variables than the only the frontal and separator union.

Affie:
Ok, 
```julia
buildSubgraphFromLabels!(dfg::AbstractDFG, cliqSubFg::AbstractDFG, frontals::Vector{Symbol},  separators::Vector{Symbol}; solvable::Int=0)
``` 
will only add the factors connected to the frontals where all the variables associated with the factor is in the clique. So you won't get the duplicates in #458.

src/SubGraphFunctions.jl
dehann: also, does this function replace existing functionality already in DFG?  buildSubGraph is a DFG thing, no?  There used to be a `buildSubGraph` in IIF, and then we moved it to DFG -- this looks like you built a new one putting it back in IIF.

@GearsAD, do you have any feelings about this in IIF.  My feeling is subGraph build functions are set-theory operations that belong in DFG?

src/SolveTree01.jl
```diff
@@ -96,6 +97,7 @@ function packFromLocalPartials!(fgl::G,
   #
 
   for idfct in getData(cliq).potentials
```
dehann: better to use the full accessor, getCliqFactors or getCliqFactorIds

src/SubGraphFunctions.jl
```diff
@@ -87,6 +87,45 @@ function buildSubgraphFromLabels(dfg::G,
...
+    if length(getNeighbors(cliqSubFg, fct)) != length(fct._variableOrderSymbols)
+      DFG.deleteFactor!(cliqSubFg, fc)
+      @error "deleteFactor! this should not happen"
+    end
+  end
+
+  return cliqSubFg
+end
+
 # function buildSubgraphFromLabels(dfg::G, syms::Vector{Symbol}) where G <: AbstractDFG
```
dehann: yep,l if memory servers this is the old one.

src/SubGraphFunctions.jl
dehann: for example these functions in DFG?
- https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/258cc40332910e97dd0ca932170d28b85e7aecbf/src/GraphsDFG/services/GraphsDFG.jl#L337
- https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/258cc40332910e97dd0ca932170d28b85e7aecbf/src/LightDFG/services/LightDFG.jl#L394

dehann:
HI @Affie , I'm concerned this is duplicate functionality that is already supposed to exist in DFG, see review comments here:
https://github.com/JuliaRobotics/IncrementalInference.jl/pull/472/files#r356738269

src/SubGraphFunctions.jl
dehann: my suggestion is to move buildSubGraphsFromLabels up to DFG

GearsAD:
I'm seeing numerical errors on the tests for this, is that a concern?

dehann:
HI @GearsAD , the numerical errors your seeing (im guessing) is on #458, which this PR is all about.  Although it has become a plumbing refactor exercise and the issue is still not resolved, but code been moved upstream to DFG (as you know) with JuliaRobotics/DistributedFactorGraphs.jl#243

src/CliqStateMachine.jl
```diff
@@ -672,8 +672,13 @@ function buildCliqSubgraph_StateMachine(csmc::CliqStateMachineContainer)
   # if isa(csmc.dfg, InMemDFGType)
   #   csmc.cliqSubFg = csmc.dfg
   # else
-    csmc.cliqSubFg = buildSubgraphFromLabels(csmc.dfg, syms, solvable=1)
+    # csmc.cliqSubFg = buildSubgraphFromLabels(csmc.dfg, syms, solvable=1)
   # end
+  frontals = getCliqFrontalVarIds(csmc.cliq)
+  separators = getCliqSeparatorVarIds(csmc.cliq)
+
+  csmc.cliqSubFg = buildSubgraphFromLabels!(csmc.dfg, csmc.cliqSubFg, frontals, separators, solvable=1)
```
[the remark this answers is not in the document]
Affie: I just quickly made the example before I left. So not intentional.

src/SolveTree01.jl
Affie: @dehann, I'm a bit concerned with this, it was a hack just to get code to run. It shouldn't be needed.

PR #490, opened 2019-12-19
Refactor of twig/test/priorusetest

GearsAD:
Updated to use https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/243

Functionality should be exactly same, small update on solvable filtering. I'm seeing numerical errors on IIF testing though, which seem to be have been on the source branch `twig/test/priorusetest` as well.

I believe `buildSubgraphFromLabels!` is functionally equivalent to `DFG._copyIntoGraph()` with a solvable filter, but we can review once this is stabilized.


dehann:
taken over by #491

Issue #495, opened 2019-12-21
[FEATURE] get variable order from tree


dehann:
getVariableOrder(tree) or getEliminationOrder(tree)

dehann:
will likely require an additional field in BayesTree structure definition.

dehann:
could probably also add the api for getVariableOrder(dfg) aswell.

dehann:
resolved by #496

PR #496, opened 2019-12-21
add variable order to tree structure

dehann:
addresses #495

### DIFF
--- a/src/JunctionTreeTypes.jl
+++ b/src/JunctionTreeTypes.jl
@@ -14,6 +14,7 @@ mutable struct BayesTree
   btid::Int
   cliques::Dict{Int,Graphs.ExVertex}
   frontals::Dict{Symbol,Int}
+  variableOrder::Vector{Symbol}
 end
 
 function emptyBayesTree()
@@ -21,7 +22,8 @@ function emptyBayesTree()
                      0,
                      Dict{Int,Graphs.ExVertex}(),
                      #[],
-                     Dict{AbstractString, Int}())
+                     Dict{AbstractString, Int}(),
+                     Symbol[] )
     return bt
 end
 
--- a/src/JunctionTree.jl
+++ b/src/JunctionTree.jl
@@ -377,7 +377,7 @@ function buildTreeFromOrdering!(dfg::G,
 
   @info "Staring the Bayes tree construction from Bayes net"
   tree = emptyBayesTree()
-  tree.variableOrdering = p
+  tree.variableOrder = p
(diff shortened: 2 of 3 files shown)


src/JunctionTreeTypes.jl
```diff
@@ -14,14 +14,16 @@ mutable struct BayesTree
   btid::Int
   cliques::Dict{Int,Graphs.ExVertex}
   frontals::Dict{Symbol,Int}
+  variableOrder::Vector{Symbol}
```
tonioteran: I see how this is adding the variableOrder member variable, but it still fails the check with an error saying that BayesTree does not have that field.

tonioteran:
Seems good to me, but for some reason I get the following error while doing a quick check---which I added as a test case:
```julia
    fg = loadCanonicalFG_Kaess()
    # Choose specific variable ordering and perform check.
    vo = [:l1, :l2, :x1, :x2, :x3]
    tree = buildTreeFromOrdering!(fg, vo)
```

accompanied by the following error:

```julia
ERROR: type BayesTree has no field variableOrdering
Stacktrace:
 [1] setproperty!(::BayesTree, ::Symbol, ::Array{Symbol,1}) at ./Base.jl:21
 [2] #buildTreeFromOrdering!#477(::Bool, ::Int64, ::Int64, ::typeof(buildTreeFromOrdering!), ::DistributedFactorGraphs.GraphsDFG{SolverParams}, ::Array{Symbol,1}) at /home/tonio/.julia/dev/IncrementalInference/src/JunctionTree.jl:380
 [3] buildTreeFromOrdering!(::DistributedFactorGraphs.GraphsDFG{SolverParams}, ::Array{Symbol,1}) at /home/tonio/.julia/dev/IncrementalInference/src/JunctionTree.jl:373
 [4] top-level scope at none:0
```

I'm not sure why it's not wanting to detect the new variable member, or if I'm missing something really silly.

dehann:
Hi @tonioteran , dang sorry i missed spelling and didn’t see this till now.  in the one place i wrote variableOrder (actual field name), but elsewhere i wrote variableOrdering.  silly mistake and lots of time, sorry about that!

dehann:
probably best to go with ‘variableOrder’ everywhere i think.

src/JunctionTreeTypes.jl
dehann: i committed a fix

tonioteran:
Can I go ahead and delete this branch?

PR #501, opened 2019-12-25
restore 458 461 test

dehann:
restores tests from #461


dehann:
okay to have large tolerance for jumpy central posterior -- just add a warning.  In time we will improve in these highly unlikely situations.

Affie:
Restore with larger test tolerances. I also previously added @test_skip and left it in place.

Issue #499, opened 2019-12-25
[REGRESSION] resolve reverse order differences for tree building


dehann:
EDIT, tasks are:
- [x] resolve double reverse
- [x] add tests to ensure both apis work as intended.

## Description

Differences when `wipeBuildNewTree!` and `resetBuildTreeFromOrdering!`

~~These two reverses are in conflict (partly introduced in #475 )~~:
- https://github.com/JuliaRobotics/IncrementalInference.jl/blob/b339a5ee2e9a01efdf63f421714d0575a7174d19/src/JunctionTree.jl#L207
- https://github.com/JuliaRobotics/IncrementalInference.jl/blob/b339a5ee2e9a01efdf63f421714d0575a7174d19/src/FactorGraph01.jl#L1098

~~Suggested refactoring is to remove the `reverse` from JunctionTree.jl Ln207~~

EDIT: First reverse in JunctionTree is related to the reverse construction order used when building the tree.  The reverse from #475 is required based on change in how QR api is used in newer versions of IIF.

cc @tonioteran

dehann:
should also add testing for both sets of API

dehann:
double reversal was not an issue, correction on this issue.  See new test in #508

Issue #464, opened 2019-12-05
how to do Gaussian belief propagation on tree


dehann:
which variables and factors (maybe conditionals) are needed in each clique?

dehann:
![b0e4bbf4-34ff-4266-9b8b-d00bc17df397](https://user-images.githubusercontent.com/6412556/70257382-2b587600-1758-11ea-9126-a2ad272ad289.jpg)

@Affie does not want to (immediately) upload himself -- so now im doing it...  Lets work from here.  Calling on @tonioteran who draws beautiful graphics.

The issue comes back to whether Bayes tree cliques are densely connected (like the Bayes Net).  Remember the Bayes tree is picked up from the Bayes Net.  Other way Bayes Net is a squashed Bayes tree.

First hypothesis is that the densely connected structure of the Bayes net is in large part carried by the ordering of the Bayes tree itself.

dehann:
variable ordering is: 0, 2, 4, 1, 3, 5

Affie:
![image](https://user-images.githubusercontent.com/6612981/70258612-2cf45f00-1795-11ea-9bb1-aacacfcc8795.png)

Affie:
![Screenshot from 2019-12-05 19-27-18](https://user-images.githubusercontent.com/6612981/70258725-6200b180-1795-11ea-984e-921e55c39671.png)

dehann:
Look, closely at the structure of the tree.  

Q: When does the relationship (conditionals) between x3, x5, x1 (root) come into affect?

A: Look at each of the children -- when calculating x3,x5 in the left child, the mixing through x4 does occur in both the up and downward direction.  

This is easy to see in a numerical approach when initial values for x3, x4, x5 exist.

Q: What if no initialization values exist?  How should information from the prior on x0 arrive at upward frontal and separators for (x4 : x3,x5).

dehann:
Also see:
- #458
- #461

dehann:
Just as a note, the "method or triples" also exists.

tonioteran:
I'll finish some graphics and then post them here to try and elucidate some concepts.

Affie:
![0E36065E-76CC-4173-92A8-F7FFB3D202DE](https://user-images.githubusercontent.com/6612981/70263143-5960a900-179e-11ea-953f-1059f19135f4.jpeg)

dehann:
[...]

PR #508, opened 2019-12-27
add few tree tests (API)

dehann:
fix for #499


dehann:
## Works

- Includes new (more comprehensive) bayes/junction tree testing as provided by the Hex example and hand tree construction of @tonioteran (#464).
- Includes testing to validate #499,
- Includes testing to validate variable order is available #495, #496 

## TODO

- These new tests do not yet include comprehensive factor clique assignment as identified by @Affie in #458 , but a (not yet fully vetted) solution was introduced in #461 (including upstream consolidation to JuliaRobotics/DistributedFactorGraphs.jl#243 and JuliaRobotics/DistributedFactorGraphs.jl#248), with other new tests pending here on #501.

Issue #481, opened 2019-12-15
[FEATURE] saveTree function


dehann:
conversion of complicated types required, see preliminaries here:
https://github.com/JuliaRobotics/Caesar.jl/blob/feat/4Q19/wtvo/examples/dev/BruteForceTreeOrderings.jl#L207-L222

PR #509, opened 2019-12-27
add save load Tree,

dehann:
addresses #481 -- still WIP

cc @tonioteran -- you might be interested in these new functions.  Still need some work before its ready for general use.


dehann:
dont merge yet, first need to finish developing loadTree

dehann:
HI @tonioteran , save and loadTree for tree only or a vector or tree is ready to use

PR #515, opened 2020-01-02
attmept lang python for dot2tex support on arm64 builds

dehann:
https://travis-ci.community/t/pip-is-not-installed-in-ppc64le-and-arm64-images/5902


dehann:
still doesnt fix the pip install issue.  Perhaps remove that requirement for arm builds.

PR #525, opened 2020-01-07
Towards breaking tree dependence on Graphs.jl

### LINKED ISSUE #500: [REFACTOR] cliq::Graphs.ExVertex -> cliq::TreeClique
## Tasks

- [x] alias for `TreeClique`
- [ ] replace `attributes` with `getAttributes`

## Description

Slow break dependence on `Graphs.ExVertex` (Graphs.jl has been archived).

Suggested solution to resolve this issue is to use a simple alias that is defined in just one place:
```julia
const TreeClique = Graphs.ExVertex
```

Simultaneously one could replace all calls `cliq.attributes` with a proper accessor `getAttributes(cliq::TreeClique)`

Affie:
fixes #500

### DIFF
--- a/src/FactorGraph01.jl
+++ b/src/FactorGraph01.jl
@@ -13,7 +13,7 @@ reshapeVec2Mat(vec::Vector, rows::Int) = reshape(vec, rows, round(Int,length(vec
 # still used for Bayes Tree
 import DistributedFactorGraphs: getData
 
-getData(v::Graphs.ExVertex) = v.attributes["data"]
+getData(v::TreeClique) = v.attributes["data"]
 
 
 """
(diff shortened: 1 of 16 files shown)


Affie:
Seems like an unrelated numerical failure only on julia 1.0

src/FactorGraph01.jl
```diff
@@ -1273,8 +1273,8 @@ function getKDE(dfg::G, lbl::Symbol) where G <: AbstractDFG
 end
 
 function expandEdgeListNeigh!(fgl::FactorGraph,
-                              vertdict::Dict{Int,Graphs.ExVertex},
-                              edgedict::Dict{Int,Graphs.Edge{Graphs.ExVertex}})
+                              vertdict::Dict{Int,TreeClique},
+                              edgedict::Dict{Int,Graphs.Edge{TreeClique}})
```
dehann: this like this will likely break when we drop Graphs.jl for tree, but thats okay for now.  Just leaving a breadcrum incase we come back here to see why things are breaking.

dehann:
@Affie, yes -- new numerical tests in IIF are still breaking here and there.  Approach is at the moment is to set tolerances quite loose.  Better solution in the future (but more Travis compute time) would be to test 10 times and allow statically as many tests to fall above or below a tighter threshold.

I will relax the offending testBasicGraphs Ln 131 slightly.  Seems like two or three of the tests are currently failing 1/10 times (current tolerances).

New tests are running now...

Issue #456, opened 2019-11-26
Provide LightGraphs alternative to Graphs.jl and remove as default.


Affie:
Graphs.jl is hard to use, slow, in julia attic, and not fully functional.
We need two parts to replace: 
- Factor Graphs, related #364 
  - [x] alternative - LightDFG
  - [x] evaluate
  - [x] swap
- BayesTree, related #365
  - [ ] alternative using LightGraphs.jl/MetaGraphs.jl see #558 
  - [ ] evaluate
  - [ ] swap

dehann:
Acknowledge and agree that we should start looking at swapping.

Affie:
Should we maybe change IIF master `InMemDFGType` to light graphs to start with the evaluation?

dehann:
Tree aspect separated andmoved to #609

Issue #365, opened 2019-08-04
Long term plan for BayesTree, probably not on Graphs.jl?


dehann:
Is there a relation between DFG and BayesTree?
- We want to be able to do BT in DB too?
- Need to work with both full clique and on clique by clique basis?

dehann:
Closing since decision is made to deprecate Graphs.jl completely.  Tree graph replacement work will be tracked via #609

PR #528, opened 2020-01-08
WIP MetaBayesTree

Affie:
Slowly removing Graphs.jl dependency and adding a MetaGraphs.jl Bayes tree. 
For first step both should work. 
Then, when we have more generic code and a better idea of how it should work, we can consider
LightBayesTree and CloudBayesTree in a similar way as with DFG.

xref #365 #456


dehann:
cc @tonioteran , just a heads up that underlying structures in BayesTree are about to change -- you will likely not see the difference when using IncrementalInference (IIF).  In case you see something related to ExVertex or TreeClique, or MetaTree show up, please let us know right away!

Affie:
Jip, i think we can rather merge sooner than later to avoid conflicts. 
The next changes becomes a bit more complicated. For now the index in metagraphs works out, but there is no clique deletion. The index needs work if we want to work towards persistent trees and tree reuse.

dehann:
Thanks!

PR #532, opened 2020-01-13
WIP MetaBayesTree in parametric

Affie:


dehann:
After thinking more, I think we should give the edge message approach a go.  If it comes up, try think on how we can share belief messages when trying to work with one clique in isolation.  Perhaps duplication of edges (when debugging individual clique solves) is totally okay.   What i mean is please add NOTES to code in case you think duplication of edges might create a problem.  We can resolve such problems if/should they come up :-)

Affie:
Hi, i don’t completely understand. How do you want to solve a clique in isolation? Do you manually send the messages? A clique will block waiting for a message to be received or send. Currently I use normal julia channels so you can just put and take from the channels without duplication needed.

dehann:
> Hi, i don’t completely understand. How do you want to solve a clique in isolation? Do you manually send the messages?

That was a bit unclear sorry -- I'm thinking of the `recordcliqs` interface of FSM which records state machine steps in the `hist` container (`::Vector`).  The dream here is to be able to store/restore each individual step of CSM with enough context to run in isolation.  For example, think about optimizing or refactoring CSM: if we can store one transition from a working sample, then we can debug and change individual steps at a time.  

My thinking around messages in cliques or edges is inspired by this "clique in isolation" or "individual CSM steps" way of thinking.  The knee jerk reaction is that a clique is a self contained unit with all necessary messages etc. -- if you have the clique at a particular point in time.  

My reservation surrounding messages on edges comes from now having to store/restore not just the clique, but also its edges.  Furthermore, the edges must now be duplicated for this or that clique, and synchronization between two cliques becomes harder and harder to track.  Hence the knee jerk.  That said, there are limits to how "defensive" the coding patterns should be.  Step by step CSM is a developers helper, but not a replacement for end-to-end testing.  Hence the relent to, just build it on edges and add NOTES.

There is a looming refactoring here on tree messages, as you already know.  This refactoring has slipped into the new year and likely best suited to do part of the *treeinit* effort.  I will likely have to punt the message refactoring another couple of weeks given my immediate schedule, but super excited to get that part cleaned up!

> A clique will block waiting for a message to be received or send.

Agreed, that sounds like the right behaviour.

> Currently I use normal julia channels so you can just put and take from the channels without duplication needed.

Agreed on Julia `::Channel`, however, the duplication I referred to above is from a different use-case origin.  While building the treeinit code (as it stands now) I was forced to add additional `::Condition` signals to the tree+CSM structure too :-P.  This has to do with cases where Siblings cannot run independent of each other any more, but need to share information through the parent, which in turn introduces the additional signaling requirements.

PR #534, opened 2020-01-15
Separate tree index and clique index

Affie:


dehann:
i'll leave this for you to merge across to other feature branch (as this is not going to master just yet)

PR #556, opened 2020-01-21
[BUGFIX] Sub-graph regression

Affie:
Related to https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/261
TODO:

- [x] In-memory types
    - GraphsDFG
    - LightDFG
- [x] Persistent types 
    - CloudDFG
    - FuseDFG
    - FileDFG
    - GraffDFG

### DIFF
--- a/src/CliqStateMachineUtils.jl
+++ b/src/CliqStateMachineUtils.jl
@@ -1045,7 +1045,7 @@ end
 function buildCliqSubgraph(fgl::AbstractDFG,
                            treel::AbstractBayesTree,
                            cliqsym::Symbol,
-                           subfg::InMemDFGType=InMemDFGType(params=getSolverParams(fgl)) )
+                           subfg::InMemoryDFGTypes=InMemDFGType(params=getSolverParams(fgl)) )
   #
   buildCliqSubgraph(fgl, treel, getCliq(treel, cliqsym), subfg)
 end


dehann:
Hi @Affie , just want to make sure we're on the same page -- the task list above should probably be expanded beyond memory and cloud.  DFG will regularly have around 5 or 6 different drivers (as per JuliaRobotics/DistributedFactorGraphs.jl#261), and there must be an implementation to move/transfer/integrate/etc between all of them.  Hence the Abstract function definitions are critical, and most effort should go towards maintaining the abstracts, because otherwise we likely to have hundreds of specialized functions which is not great from maintenance standpoint.

The main use-case of DFG is not a one-to-one or silo'ed wrapper of outside graphing libraries.  The main use case of DFG is a standard mechanism to work with data between/to/from any of these technologies (the harder problem).  With JuliaRobotics/DistributedFactorGraphs.jl#262 we tried to flatten that out a bit.

If there really really really is a case where the abstract cannot be made good enough then we can look to add a specialized function, but someone will have to own that and jump on for all cases where it might be used.  We should really try work at maintaining a common definitions.

Affie:
:+1: Edited first post with your recommendation.

Affie:
I think rest is fixed in https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/263. Open new issue if something fails

PR #572, opened 2020-02-03
fix getFactorDim return as Int

dehann:


dehann:
follows from error in #573

PR #583, opened 2020-02-11
fix some deprecate warnings for v0.6

Affie:


test/testCompareVariablesFactors.jl
```diff
@@ -90,8 +90,8 @@ addFactor!(fg, [:x1;:l1], LinearConditional(Rayleigh()))
 
 sfg = buildSubgraphFromLabels!(fg, [:x0;:x1])
 
-
-@test compareFactorGraphs(fg, sfg, skip=[:labelDict;:addHistory;:logpath])
+#FIXME JT - this doesn't make sense to pass?
+# @test compareFactorGraphs(fg, sfg, skip=[:labelDict;:addHistory;:logpath])
```
Affie: this doesn't make sense to pass

src/FGOSUtils.jl
```diff
@@ -25,8 +25,9 @@ Related
```
dehann: should we delete the docs here too?

src/JunctionTree.jl
```diff
@@ -1633,7 +1633,7 @@ end
 
 Return the variable order stored in a tree object.
 """
-getVariableOrder(treel::AbstractBayesTree)::Vector{Symbol} = treel.variableOrder
+DistributedFactorGraphs.getVariableOrder(treel::AbstractBayesTree)::Vector{Symbol} = treel.variableOrder
```
dehann: tree not defined in DFG, is it?  Or is this a function name override issue?

If so, we should rather import the name from DFG and override here for dispatch.  I'm a little confused on why calling DFG explicitly here is the right thing to do.

test/testCompareVariablesFactors.jl
dehann: okay, that sounds fine

Issue #568, opened 2020-01-30
[BUG] approxConv is not updating ccw.measurement in calling function


dehann:
```julia
fg = ...
fcm = solverData(getFactor(fg, :x1x2f1)).fnc.measurement |> deepcopy
approxConv(fg, :x1x2f1, :x1)
fcm2 = solverData(getFactor(fg, :x1x2f1)).fnc.measurement
```

and the values in fcm's are the same!  The factor `ccw.measurements` location should have been updated by the internal `evalFunctionSpecific` call.  Quick testing showed that the internal getSample method is called and `ccwl.measurements` is updated, but somehow that data is lost on the way out as `approxConv` finishes.

-----

These were somewhat useful during first finding this bug:
```julia
solverData(getFactor(fg, :x1x2f1)).fnc |> pointer_from_objref
solverData(getFactor(fg, :x1x2f1)).fnc.measurement[1] |> pointer_from_objref
```

dehann:
also
```juila
  @show pointer_from_objref(ccw)
  @show (solverData(fct).fnc |> pointer_from_objref)
  @show solverData(fct).fnc.measurement[1] |> pointer_from_objref
```

dehann:
Looks like this was resolved as part of DFG v0.6, IIF v0.9, and #639

PR #588, opened 2020-02-12
Development updates for DFG v0.6

Affie:


src/CliqStateMachineUtils.jl
```diff
@@ -898,7 +898,8 @@ directPriorMsgIDs, directFrtlMsgIDs, directAssignmentIDs, mcmcIterationIDs
 function determineCliqVariableDownSequence(subfg::AbstractDFG, cliq::TreeClique; solvable::Int=1)
   frtl = getCliqFrontalVarIds(cliq)
 
-  adj = DFG.getAdjacencyMatrix(subfg, solvable=solvable)
+  #TODO don't use this getAdjacencyMatrixSymbols
+  adj = DFG.getAdjacencyMatrixSymbols(subfg, solvable=solvable)
```
Affie: This has to be updated to use `getBiadjacencyMatrix`

src/FactorGraph01.jl
```diff
@@ -552,7 +548,7 @@ end
 
 Fetch the variable marginal sample points without the KDE bandwidth parameter.  Use getVertKDE to retrieve the full KDE object.
 """
-function getVal(vA::Vector{DFGVariable}, solveKey::Symbol=:default)::Array{Float64, 2}
+function getVal(vA::Vector{<:DFGVariable}, solveKey::Symbol=:default)::Array{Float64, 2}
```
Affie: This is because a DFGVariable now takes type parameters for softtype. like DFGFactor

src/FactorGraph01.jl
```diff
@@ -1017,19 +1015,21 @@ function addFactor!(dfg::G,
...
+  newFactor = DFGFactor(Symbol(namestring),
+                        varOrderLabels,
+                        solverData;
+                        tags=Set(union(labels, [:FACTOR])),
+                        solvable=solvable)
 
   # TODO: Need to remove this...
   for vert in Xi
-    push!(newData.fncargvID, vert.label) # vert._internalId # YUCK :/ -- Yup, this is a problem
+    push!(solverData.fncargvID, vert.label)
```
Affie: I tried to follow through on @GearsAD changes,. Please just confirm if this is what you wanted.

src/FactorGraph01.jl
```diff
@@ -1017,19 +1015,21 @@ function addFactor!(dfg::G,
...
   end
 
-  success = DFG.addFactor!(dfg, Xi, newFactor)
+  success = DFG.addFactor!(dfg, newFactor)
```
Affie: Note, I recommend we use this with variable order extracted from the factor.

src/DispatchPackedConversions.jl
```diff
@@ -135,20 +135,20 @@ completely rebuild the factor's CCW and user data.
 function rebuildFactorMetadata!(dfg::G, factor::DFGFactor)::DFGFactor where G <: AbstractDFG
   # Set up the neighbor data
   neighbors = map(vId->getVariable(dfg, vId), getNeighbors(dfg, factor))
-  neighborUserData = map(v->solverData(v).softtype, neighbors)
+  neighborUserData = map(v->getSolverData(v).softtype, neighbors)
 
   # Rebuilding the CCW
-  setDefaultFactorNode!(dfg, factor, neighbors, factor.data.fnc.usrfnc!)
+  ccw_new = getDefaultFactorData(dfg, neighbors, factor.data.fnc.usrfnc!)
+  setSolverData!(factor, ccw_new)
```
Affie: Also confirm this.

src/CliqStateMachineUtils.jl
dehann: understood

src/DispatchPackedConversions.jl
dehann: i recently found this issue, might be somewhat related to your concern here: #568

src/FactorGraph01.jl
dehann: sounds good!

src/FactorGraph01.jl
dehann: defer this to @GearsAD please

src/FactorGraph01.jl
dehann: Perhaps I could ask for a bit more detail on this example please?  Willing to give it a try, as long as we provide both APIs until we decide which is the long term strategy.

Affie: Both can stay forever. just recommend this one. So drivers should implement this one, the others are all done for AbstractDFG.

Affie: see here: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/3ec2484a2b22a00b47027a310e284ff674f08a9c/src/services/AbstractDFG.jl#L217-L236

PR #600, opened 2020-02-25
Add covariance calculations to parametric

### LINKED ISSUE #539: Calculate covariances for parametric tree solve
It will not be implemented in PR #468 and is needed to compete the tree solve.

Affie:
part of #539 
We need new messages before covariances can be used in the tree, related #459


Affie:
I split the parts that are more or less working for a smaller PR. 
- Covarariance calculation is not 100% yet.  
- Treesolve only uses priors.

PR #626, opened 2020-03-01
Fixing DiagNormal

GearsAD:


GearsAD:
Sqrt because of https://github.com/JuliaStats/Distributions.jl/issues/584.

TODO: Cleanup when paper is in!

PR #650, opened 2020-03-19
fix tests after prev merge

dehann:
part of resolving #459


dehann:
FYI @Affie, changing to `addMsgFactors_Parametric!`

PR #652, opened 2020-03-20
fast forward parametric dev

dehann:
FYI, @Affie note Project.toml at v0.11.0 but not tagged yet.  I have been moving code around and fixing bugs we missed yesterday.  I'd suggest bringing your feature branches forward as soon as possible.

Idea of the new work in master is to better prepare for consolidation work in #459 

PS, there is another similar branch here on IIF with almost exactly the same name as this feature branch?


dehann:
PS, there is another similar branch here on IIF with almost exactly the same name as this feature branch?

Issue #578, opened 2020-02-05
errors on large batchSolve


dehann:
```julia
Base.InvalidCharError{Char}('\xee')
Stacktrace:
 [1] invalid_char(::Char) at ./char.jl:85
 [2] Type at ./char.jl:130 [inlined]
 [3] convert at ./char.jl:180 [inlined]
 [4] cconvert at ./essentials.jl:391 [inlined]
 [5] is_id_char(::Char) at ./show.jl:792
 [6] _all(::typeof(Base.is_id_char), ::Base.Iterators.Rest{String,Int64}, ::Colon) at ./reduce.jl:737
 [7] all at ./reduce.jl:732 [inlined]
 [8] isidentifier(::String) at ./show.jl:797
 [9] show_unquoted_quote_expr(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any, ::Int64, ::Int64) at ./show.jl:1047
 [10] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Symbol) at ./show.jl:733
 [11] _show_default(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:347
 [12] show_default at ./show.jl:330 [inlined]
 [13] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:327
 [14] _show_default(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:347
 [15] show_default at ./show.jl:330 [inlined]
 [16] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at ./show.jl:327
 [17] _show_default(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Any) at ./show.jl:347
 [18] show_default at ./show.jl:330 [inlined]
 [19] show at ./show.jl:327 [inlined]
 [20] print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::RemoteException) at ./strings/io.jl:37
 [21] print_to_string(::RemoteException) at ./strings/io.jl:129
 [22] string at ./strings/io.jl:168 [inlined]
[...]

PR #658, opened 2020-03-24
several bugfixes

dehann:


dehann:
likely address issues related to #578

PR #672, opened 2020-04-04
drawtree in loop as standard

dehann:


Affie:
This is good, I was considering something similar with the addition of using a clique status to set the clique colour.

Issue #373, opened 2019-08-15
make sure bss sampling of Uniform without NaN is working properly


dehann:
near about IIF#2522145 seems an issue was introduced to `AliasingScalarSampler`

PR #679, opened 2020-04-09
Consolidate Subgraph functions (DFG #373) 

Affie:


dehann:
this needs DFG v0.7.3

Issue #615, opened 2020-02-29
Reinitialize functionality and storing initialized state option in supersolve


Affie:
Hacky way of "resetting" initialization.
```julia
## Decompose and rebuild fg

newfg = LightDFG{SolverParams}(params=SolverParams(algorithms=[:default, :parametric]))

for v in getVariables(fg)
  addVariable!(newfg, v.label, getSofttype(v))
end

flist = sortDFG(lsf(fg))

for fl in flist
  f = getFactor(fg, fl)
  addFactor!(newfg, getVariableOrder(f), getFactorFunction(f), autoinit=true)
end
```

dehann:
related to #612

dehann:
```julia
fgNew = deepcopy(fg)
resetInitialValues!(fgNew,fg)
```

dehann:
`resetVariableToInitial!` might be a good alias too.

Issue #612, opened 2020-02-28
add :graphinit as solvekey to supersolves


dehann:
```julia
julia> whichSolutions(fg)
:default
:parametric
:graphinit
:treeinit
:reference
```

PR #688, opened 2020-04-22
Additional updates to persist the values on initialization

GearsAD:
On initialization, persist the values - these should be no-ops for in-memory.

### DIFF
--- a/src/FactorGraph.jl
+++ b/src/FactorGraph.jl
@@ -824,8 +824,10 @@ function doautoinit!(dfg::T,
         end
         pts,inferdim = predictbelief(dfg, vsym, useinitfct, logger=logger)
         setValKDE!(xi, pts, true, inferdim)
-        #TODO test
-        setVariablePosteriorEstimates!(xi)
+        # Update the estimates (longer DFG function used so cloud is also updated)
+        setVariablePosteriorEstimates!(dfg, xi.label)
+        # Update the data in the event that it's not local
+        mergeVariableSolverData!(dfg, xi)
         didinit = true
       end
     end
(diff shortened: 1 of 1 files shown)


src/FactorGraph.jl
```diff
@@ -824,8 +824,10 @@ function doautoinit!(dfg::T,
         end
         pts,inferdim = predictbelief(dfg, vsym, useinitfct, logger=logger)
         setValKDE!(xi, pts, true, inferdim)
-        #TODO test
-        setVariablePosteriorEstimates!(xi)
+        # Update the estimates (longer DFG function used so cloud is also updated)
+        setVariablePosteriorEstimates!(dfg, xi.label)
+        # Update the data in the event that it's not local
+        mergeVariableSolverData!(dfg, xi)
```
Affie, line 830: Perhaps this should rather be updateVariableSolverData! with a specific key. I’ll have to look in more detail, but merge didn’t work for the solved update as it could overwrite other keys.

Just as a breadcrumb, addSolverData(dfg, xi, :graphinit) for #615

GearsAD: Okay, so I went through https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/wiki/Standard-Definitions-of-API-via-verbs-and-set-theory and looked at the functions (think they're still in flux), but I couldn't see the distinction between the two. Yes i think it should update just the initialized solver data, but it seems that the init method always just initializes the default key? Will update, but think we should clear up the distinctions

Affie: Yes, I think currently it shouldn't make a difference.

dehann: I agree with that this should be solveKey specific, since this is part of the current nonparametric graph initialization code.  Could we swap to verb `update` and ensure `solveKey=:default`.

I see verbNoun for `dest = merge*(src1,src2)` can be cleaned up a bit, and understand `merge` as creating a third collection but `merge*!(dest, src)` overwrites the first (the Julian way).

dehann: @GearsAD , I made the change -- did you want to check it out before merging?

dehann:
oh dont merge yet -- its not quite working yet.  testing locally now

dehann:
part of #612 #615

PR #695, opened 2020-05-05
Develop [ing something??]

dehann:


dehann:
if DFG v0.8.0 has a breaking change, we should perhaps lift the DFG minimum requirement to v0.8.0?

Affie:
Minor version will need to be bumped.

dehann:
now requires JuliaRobotics/DistributedFactorGraphs.jl#442 for IIF tests to pass

dehann:
help fix JuliaRobotics/DistributedFactorGraphs.jl#441

dehann:
> all will use the same wrapper.

we all using CCW and that keeps the code maintenance down too :-)

Affie:
> if DFG v0.8.0 has a breaking change, we should perhaps lift the DFG minimum requirement to v0.8.0?

I had to put v0.7 back because of a weird travis issue, but yes. It should be removed

PR #700, opened 2020-05-07
Also dispatch on solverData type for rebuildFactorMetadata!

Affie:
to match DFG #449


dehann:
Hi @Affie , just checking if you meant to break the build here -- `Type` is not `UnionAll` ?
https://travis-ci.org/github/JuliaRobotics/IncrementalInference.jl/jobs/684173072

dehann:
Ah, never mind -- I was missing DFG master code on my local machine.  Thanks!

PR #709, opened 2020-05-23
standardize to use solverParams and not params

Affie:
see [DFG#475](https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/475)


dehann:
thanks, will merge after this: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/475#issuecomment-633187166

PR #722, opened 2020-06-05
add manifold and per particle nullhypo

### LINKED ISSUE #713: multihypo spreadNH should add entropy per particle
Not add entropy around mean of particles.  From issues reported during use where two modes are observed to merge together too easily after #661.

Related to #664, #248

dehann:
fix #713

Still have to generalize for common nullhypo usage, but this is a step in the right direction.


dehann:
bad branch name sorry

PR #732, opened 2020-06-10
compat julia 1.4

dehann:


.travis.yml
```diff
@@ -27,7 +27,6 @@ jobs:
   allow_failures:
     - julia: nightly
     - arch: arm64
-    - name: "MASTER AND TAGS"
```
Affie, line 30: @dehann Why don’t you like this?

dehann: Idea was to keep ARM as allow failure, but then see how far we get with the usual MASTER AND TAGS as required test.  DFG v0.8 was a big lift.  MASTER AND TAGS should be working at this time, right?

PR #757, opened 2020-06-25
Update travis - allow failures on ARM

### LINKED ISSUE #755: Travis bug with ARM allow failures
Just escalating the review to an issue so it doesn't get missed.
The bug is caused by this deletion:

@dehann Why don’t you like this?

_Originally posted by @Affie in https://github.com/JuliaRobotics/IncrementalInference.jl/pull/732_

it makes this line incorrect:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/03d231d63fad0f7b2ea897c328a657753ca8d7c2/.travis.yml#L30

Affie:
Close #755

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,9 @@ jobs:
   allow_failures:
     - julia: nightly
     - arch: arm64
-      if: NOT branch =~ ^release.*$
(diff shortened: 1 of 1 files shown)


dehann:
HI @Affie, thanks!

Issue #443, opened 2019-11-20
Documenting CSM and inference steps for debugging


dehann:
### Console

```julia
# save clique subgraphs during inference at critical steps
getSolverParams(fg).dbg = true
# record CSM steps for later analysis
tree, smt, hist = solveTree!(fg, recordcliqs=ls(fg))

# print to screen the CSM steps for clique with frontal :x2
printCliqHistorySummary(tree, :x2)

# where is everything saved???
getLogPath(fg)

clid = getCliq(tree, :x5).index
STEP = 8
fncAS = hist[clid][STEP][3]
dataAS = deepcopy(hist[clid][STEP][4])
fncAS(dataAS)
```

- `solveTree!` will always generate content at `getLogPath(fg)` location.  
- More or less will be stored there depending `getSolverParams(fg).dbg`.
- `recordcliqs=[:x1;:x2;...]` will save the CSM history to RAM.

### Animate CSM Video

Can also make concurrent video of CSM at work:
```julia
csmAnimate(fg,tree,ls(fg), frames=1000)

run(`ffmpeg -r 10 -i /tmp/caesar/csmCompound/csm_%d.png -c:v libtheora -vf fps=25 -pix_fmt yuv420p -vf "scale=trunc(iw/2)*2:trunc(ih/2)*2" -q 10 /tmp/caesar/csmCompound/out.ogv`)

@async run(`totem /tmp/caesar/csmCompound/out.ogv`)
```

Also see `csmAnimate`'s own documentation.

dehann:
cc @tonioteran @Affie

dehann:
NOte that when using `getSolverParams(fg).dbg=true` each of the major stages of the CSM clique inference subgraphs are stored to the `getLogPath(fg)` destination -- including:
- `fg_build`, as copied from main fg object
- `fg_beforeupsolve`
- `fg_afterupsolve`
- `fg_beforedownsolve`
- `fg_afterdownsolve`

which shows the belief messages passed up or down in each case added as `MsgPriors` to the clique subgraphs.

Affie:
Is there a way to debug with a failed solve (where it errors and doesn't return)?

dehann:
Hi @Affie , yeah so the best I have been able to do is as follows:

First thing is to find where the logs are going to land:
```julia
@show getLogPath(fg)
```

Then solve with all debug options:
```julia
getSolverParams(fg).dbg = true
getSolverParams(fg).async = true

[...]

PR #758, opened 2020-06-25
WIP, Monitor parametric csms and cleanly exit on fail

### LINKED ISSUE #753: add watchdog to auto-stop dormant CSM tasks
If during `solveTree!` something goes wrong and there is no activity anywhere on the tree (all CSMs are stuck or closed out) then an autostop on all tasks  -- perhaps self terminate on watchdog -- could be very useful, rather than having to close the entire Julia context to regain control of threads/processes.

Affie:
@dehann, this got left behind somehow. I'll wait for #459 and then see if it still works. 
it was my solution to #753

For future self - look at local changes in that branch

### DIFF
--- a/src/ParametricSolveTree.jl
+++ b/src/ParametricSolveTree.jl
@@ -28,6 +28,7 @@ function taskSolveTreeParametric!(dfg::AbstractDFG,
   cliqHistories = Dict{Int,Vector{Tuple{DateTime, Int, Function, CliqStateMachineContainer}}}()
   if !isTreeSolved(treel, skipinitialized=true)
     @sync begin
+      monitortask = monitorCSMs(treel, alltasks)
       # duplicate int i into async (important for concurrency)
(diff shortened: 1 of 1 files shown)


dehann:
Will ask this to be added for the nonparametric case soon too please :-) almost there on 459...

dehann:
Ah also, this is not a watchdog timer in the same sense -- I read this as termination when one CSM fails.  Idea in #753 also includes a further case where a CSM goes unresponsive without an error occurring, where a watchdog then interrupts and terminates the entire solve is this PR is doing.

Affie:
> this is not a watchdog timer in the same sense -- I read this as termination when one CSM fails

Yes, Its clean termination on fail mechanism. I terminated manually so far on stalls. I went for the global approach to monitoring and not in the cliques, but the design is not completed yet.

dehann:
xref: https://github.com/JuliaRobotics/IncrementalInference.jl/issues/443#issuecomment-677702617

PR #781, opened 2020-07-10
Up msg consolidation #459 done

dehann:
up part of #459 done.  Note testing in RoME continues to verify stability.

### DIFF
--- a/src/Deprecated.jl
+++ b/src/Deprecated.jl
@@ -5,6 +5,7 @@
 ## Delete at end v0.13.x
 ##==============================================================================
 
+export setCliqueMsgUp!
 export getMsgUpInitChannel_
 export putMsgUpInit!, fetchMsgUpInit
 export putMsgUpInitStatus!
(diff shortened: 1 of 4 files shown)


dehann:
So check this out:  To do Hex with graphinit:
```julia
using RoME, RoMEPlotting

fg = generateCanonicalFG_Hexagonal()
solveTree!(fg)
plotSLAM2D(fg)
```

## To use treeinit instead:
```julia
fg = generateCanonicalFG_Hexagonal(graphinit=false)
getSolverParams(fg).graphinit = false
getSolverParams(fg).treeinit = true
solveTree!(fg)
plotSLAM2D(fg)
```
Booyah!
Screenshot from 2020-07-10 17-52-12


## Notes on This PR for #459 

This PR concludes the upward message consolidation part of #459, but the work remains to consolidate down message infrastructure in the same way.  Note that this up message structure can be used as a template/guide for down message work.

Note for up messages that only one Channel container exists, that is `BTND.upMsgChannel::Channel{LikelihoodMessage}` -- nothing else.  Accessors are `put!` and non-destructive `fetch` -- i.e. there is no user `take!` operation on the message channel.

## Next Steps

Hi @Affie, would you be willing to pick up down message portion of the consolidation from here please?  Lets keep channels in BTND for now please and not on the tree edges.  Since its only a single message channel, we can move that later quite easily.  Also, you'll note that ParametricCSM is just a little behind on API, but should all be very close now.  Main action is the function `TreeMessageAccessors.jl:prepPutCliqueStatusMsgUp!`.

cc @GearsAD

PR #794, opened 2020-07-19
Renaming factor base types

GearsAD:
Merge after https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/537

### DIFF
--- a/src/ApproxConv.jl
+++ b/src/ApproxConv.jl
@@ -12,7 +12,7 @@ Future work:
 
 """
 function approxConvOnElements!(ccwl::CommonConvWrapper{T},
-                               elements::Union{Vector{Int}, UnitRange{Int}}, ::Type{MultiThreaded}  ) where {T <: Union{FunctorPairwise, FunctorPairwiseMinimize}}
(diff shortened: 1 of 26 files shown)


src/ApproxConv.jl
```diff
@@ -252,12 +252,12 @@ end
 """
     $(SIGNATURES)
 
-Do true and null hypothesis computations based on data structures prepared earlier -- specific to `FunctorPairwiseNH`.  This function will be merged into a standard case for `FunctorPairwise/Minimize` in the future.
+Do true and null hypothesis computations based on data structures prepared earlier -- specific to `AbstractRelativeFactorNH`.  This function will be merged into a standard case for `AbstractRelativeFactor/Minimize` in the future.
```
dehann, line 255: `FunctorPairwiseNH` is about to deprecated anyway, so we can leave these.

dehann: part of the work to close #237

GearsAD:
Ah, I think that's used in RoME though?

GearsAD:
I'll revert the RoME changes for NH

PR #798, opened 2020-07-20
UTC DateTime to ZonedDateTime localzone() 

Affie:
and fix  LinearConditional() warning.

> NOTE: Requires DFG 0.9


src/DefaultNodeTypes.jl
```diff
@@ -169,7 +169,7 @@ Default linear offset between two scalar variables.
 struct LinearConditional{T} <: AbstractRelativeFactor where T <: SamplableBelief
   Z::T
 end
-LinearConditional(z::T=Normal()) where T = LinearConditional{T}(z)
+LinearConditional() = LinearConditional(Normal())
```
dehann, line 172: Assuming then that new Julia automatically produces the helper functions?
```julia
LinearConditional(t::T) where T = LinearConditional{T)(t)
```

Affie: yes, julia creates an inner and outer constructor automatically if none are provided:
>Without any explicitly provided inner constructors, the declaration of the composite type Point{T<:Real} automatically provides an inner constructor, Point{T}, for each possible type T<:Real, that behaves just like non-parametric default inner constructors do. It also provides a single general outer Point constructor that takes pairs of real arguments, which must be of the same type.

That's what I have against providing inner constructors when it is not needed, its a lot of maintenance and it's unnecessary most of the time.

So `LinearConditional(z::T=Normal()) where T = LinearConditional{T}(z)` overwrote the default outer constructor leading to a warning and breaking incremental compilation.

Also its a bit strange to have a default `LinearConditional()`

PR #814, opened 2020-08-02
deconv msg likelihoods option

### LINKED ISSUE #577: Ensure clique message priors calculation captures multivariate correlations
Similar discussion to #464.

When calculation parent belief say X1,X2,X3 ; with joint upward prior on X1, X3-- make sure calculations on X1 and X3 in parent (which happen consecutively) maintains the correlation between samples.  I.e. when solving particles 1..N, don't resample separately.

dehann:
- partially resolve #577 ;
- only active with `opt.useMsgLikelihoods` at this time;
- effects both upward and downward solve.


dehann:
[test577.pdf](https://github.com/JuliaRobotics/IncrementalInference.jl/files/5013079/test577.pdf)

PR #819, opened 2020-08-04
Softtype - use getDimension and getManifolds 

Affie:


dehann:
HI @Affie , you can merge when ready thanks!

PR #823, opened 2020-08-05
Populate full variable and use ns time

Affie:


src/ApproxConv.jl
```diff
@@ -103,6 +103,8 @@ function prepareCommonConvWrapper!(ccwl::CommonConvWrapper{T},
     ccwl.cpt[i].p = collect(1:size(ccwl.cpt[i].X,1)) # collect(1:length(ccwl.cpt[i].Y))
     ccwl.cpt[i].Y = zeros(ccwl.xDim)  # zeros(ccwl.partial ? length(ccwl.usrfnc!.partial) : ccwl.xDim )
     ccwl.cpt[i].res = zeros(ccwl.xDim) # used in ccw functor for AbstractRelativeFactorMinimize
+    # TODO JT - Confirm it should be updated here.
+    ccwl.cpt[i].factormetadata.fullvariables = copy(Xi)
```
dehann, line 107: is a copy necessary?  Factor operations should not be allowed to change the variable metadata?

src/FactorGraph.jl
```diff
@@ -398,19 +398,32 @@ function addVariable!(dfg::AbstractDFG,
...
                       checkduplicates::Bool=true,
                       initsolvekeys::Vector{Symbol}=getSolverParams(dfg).algorithms)::DFGVariable
   #
+
+  if :ut in fieldnames(typeof(softtype))
+    Base.depwarn("Field `ut` (microseconds) for softtype $(softtype) has been deprecated please use DFGVariable.nstime, kwarg: nanosecondtime", :addVariable!)
+    if isnothing(nanosecondtime)
+      nanosecondtime = Nanosecond(softtype.ut*1000)
+    else 
+      @warn "Nanosecond time has been speciefied as $nanosecondtime, ignoring `ut` field value: $(softtype.ut)."
```
dehann, line 415: speciefied -> specified

src/ApproxConv.jl
Affie: Copy was to be on the save side. I want to be sure the container does not change. It looks like it doesn't, but there are some comments in the code about constructing `Xi`:
> # TODO -- this build up of Xi is excessive and could happen at addFactor time
Perhaps `fullvairables` can be populated at construction time?

src/FactorGraph.jl
Affie: thanks, missed that

src/ApproxConv.jl
Affie: @dehann perhaps rather in `prepgenericconvolution`? Or will it not make a difference.

src/FactorGraph.jl
```diff
@@ -608,6 +621,8 @@ function prepgenericconvolution(
         )
   #
   for i in 1:Threads.nthreads()
+    # TODO JT - Confirm it should be updated here. Also testing in prepareCommonConvWrapper!
+    ccw.cpt[i].factormetadata.fullvariables = copy(Xi)
```
dehann, line 625: again, not sure if the copy is really needed?

PR #828, opened 2020-08-12
Cleanup for DFG v.10.0

GearsAD:


src/Deprecated.jl
```diff
@@ -27,8 +27,8 @@ end
 
 
 export Pairwise, Singleton
-export FunctorSingletonNH, FunctorPairwiseNH
-export FunctorInferenceType, FunctorPairwise, FunctorPairwiseMinimize, FunctorSingleton
+export AbstractPriorNH, AbstractRelativeFactorNH
```
dehann, line 30: These __NH items can be dropped entirely.  IIF v0.14 has standardized the `nullhypo` approach, so this is no longer required

PR #832, opened 2020-08-13
clean up addVariable!

dehann:
partially addresses JuliaRobotics/DistributedFactorGraphs.jl#603

### DIFF
--- a/src/FactorGraph.jl
+++ b/src/FactorGraph.jl
@@ -381,7 +381,11 @@ end
 """
 $(SIGNATURES)
 
-Add a variable node `lbl::Symbol` to `fg::AbstractDFG`, as `softtype<:InferenceVariable`.
+Add a variable node `label::Symbol` to `dfg::AbstractDFG`, as `varType<:InferenceVariable`.
+
+Notes
+-----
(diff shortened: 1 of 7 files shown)


src/FactorGraph.jl
```diff
@@ -392,84 +396,67 @@ addVariable!(fg, :x0, Pose2)
...
-  if :ut in fieldnames(typeof(softtype))
-    Base.depwarn("Field `ut` (microseconds) for softtype $(softtype) has been deprecated please use DFGVariable.nstime, kwarg: nanosecondtime", :addVariable!)
-    if isnothing(nanosecondtime)
-      nanosecondtime = Nanosecond(softtype.ut*1000)
-    else 
-      @warn "Nanosecond time has been specified as $nanosecondtime, ignoring `ut` field value: $(softtype.ut)."
-    end
-  elseif isnothing(nanosecondtime)
-    nanosecondtime = Nanosecond(0)
-  end
```
Affie, line 419: @dehann do you have a reason for deleting this? Or can I put it back?

dehann: oh sorry, that was probably an accident!  I was clearing out lots of really old code and this was probably taken with.  We can put it back.

PR #833, opened 2020-08-14
remove import loadDFG

Affie:


src/FGOSUtils.jl
```diff
@@ -2,7 +2,7 @@
 #  IIF methods should direclty detect extended types from user import
 # of convert in their namespace
 
-import DistributedFactorGraphs: AbstractPointParametricEst, loadDFG
```
dehann, line 5: I just added loadDFG yesterday?  What's the issue?

Affie: It's not in DFG to import

dehann: https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/60ba70fdf026d995e727f72ef021db10835a5c2e/src/FileDFG/services/FileDFG.jl#L171

dehann: https://github.com/JuliaRobotics/IncrementalInference.jl/blob/94bc84fd3c22c9d1c8ef09410bd58409c5efc681/src/FGOSUtils.jl#L358

Affie: Haha, sneaky. I updated IIF, but DFG was slightly behind.

PR #839, opened 2020-08-17
mmd analysis tools for better solution validation

dehann:
tools to
- draw mmd for consecutive solve change in each/all variables,
- draw mmd for latest solve against init value for each/all variables,
- draw mmd for consecutive solve change in each/all factors (deconv),
- draw mmd for latest solve against meas value for each/all factors.


dehann:
cc @Affie , we spoke about this a little the other day.  Just wanted to add you in here incase you wanted to follow along.

dehann:
important info transferred to #1363

PR #840, opened 2020-08-17
Put back softtype (variableType) deprecations simplify addVariable!

Affie:
removed in: https://github.com/JuliaRobotics/IncrementalInference.jl/pull/832

should close https://github.com/JuliaRobotics/RoME.jl/issues/324

### DIFF
--- a/src/FactorGraph.jl
+++ b/src/FactorGraph.jl
@@ -378,6 +378,10 @@ function setVariableRefence!(dfg::AbstractDFG,
 end
 
 
+# get instance from variableType
+_variableType(varType::InferenceVariable) = varType
+_variableType(varType::Type{<:InferenceVariable}) = varType()
+
 """
 $(SIGNATURES)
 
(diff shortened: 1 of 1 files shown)


src/FactorGraph.jl
```diff
@@ -397,18 +401,32 @@ addVariable!(fg, :x0, Pose2)
 """
 function addVariable!(dfg::AbstractDFG,
                       label::Symbol,
-                      varType::InferenceVariable;
+                      varTypeU::Union{T, Type{T}};
```
dehann, line 404: Cool thanks, I created a new const called `InstanceType{T}`:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/05d5d732fff020c5c4b7aad7b0813735bccc13a9/src/IncrementalInference.jl#L72

Affie: I saw that one. I didn't think it would work as T kan be of type `InferenceVariable` and then you would get a method error in the softtype constructor and not in `addVarialbe!`. (in `varType()`)

Issue #855, opened 2020-08-24
fetch vs take! on message channels


Affie:
Currently parametric works with a take! and save (for later use and debugging) model. ~~(or used to, it's broken)~~ restored with #857

CSM works with `fetch` and `conditions` with channels stored in the source clique ("pull")

Part of #459

Affie:
The "take! and save" approach in parametric is based on:
- The "4-stroke" state machine with messages going all the way for initialization also (leaves <--> root)
    - Based on asynchronous vs synchronous message passing algorithms.
- Therefore, cliques only calculate messages once per stroke.
- I wanted to rely on the **channels** alone for **synchronisation** and wait for all children to send a message.

dehann:
Now that i'm thinking about 4-stroke more, I can think of one case that might be tricky to do without "cascading", but lets worry about that after consolidation.

dehann:
> synchronisation I also mean to wait on

For tree init the status needs to be checked multiple times, and later messages need to be fetched more than once -- this make the take! on one channel only harder i think.  I was not able to resolve tree init without switching to a fetch model with Condition to allow a wait cycle again.

>  channels are basically just wrappers for convenience around conditions (with a buffer)

Yes and no, `Conditions` are edge triggered while `Channels` are level triggered.  It's discussed somewhere in the Julia docs.

You'll see in CSM there are states that are `slowXYZ`.  This is usually a wait on condition with checks for some kind of status, and also short loops in the CSM should a certain criteria be met.  This is sometimes required for treeinit where a hard block until upsolved is not always possible, or when siblings need to decide on a down init order (thats in the cascading model).

Affie:
> Yes and no, Conditions are edge triggered while Channels are level triggered

Yes, you are correct. For me, that is an advantage, with level triggering cliques don't need to be waiting when the message is sent.

dehann:
Just confirming, for the take model you send a copy of the down message to each child?

Affie:
Yes. As you know there is a channel per edge. Currently for down the same message is sent on all down channels. I considered a possible future need to send a specific message to only one child, but that didn’t seem necessary yet.

Affie:
The cliques are scheduled by the repeating pattern:
- ....
- send up
- wait for down
- do some down stuff
- send down
- wait for up
- do some up stuff
- ....

Affie:
WIP take! storyboard
[...]

Issue #675, opened 2020-04-04
Decide if msg channels in BTND or edges only


dehann:
EDIT
 
## CONCLUSION

https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-676488463

-----

## Original Post

Following from #459, we should decide if message channels between cliques are stored in either csmc or tree edges.

Related decisions:
- [x] #674 (~~push or~~ pull model)
- [ ] #673 (simultaneous solvers need separate channels)

Part of the decision is whether the tree should be read-only during inference.

---

WIP:
- There are strong debug-tool reasons to store msgs in tree, unless messages are extracted and separately available for debugging (https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-609013503)

dehann:
I have a preference for storing message channels inside CSMCs rather than on tree edges, loosely discussed in #459.

dehann:
functions like `treeProductUp` do not have access to CSMC, since this is a debugging function that occurs separate of solver (nominally post-inference to debug a weird internal result).

Affie:
We should maybe make the distinction between the message channels and the messages itself. 
Message storage options are in clique or CSMC. 
For me, the message channels only makes sense in the tree or cliques (with my preference in the tree as you know), not in CSMC.

dehann:
> (with my preference in the tree as you know), not in CSMC.

Ah, I had it that your preference was not in the tree, maybe make the tree read only?

Nonparametric is currently storing the message channels in the tree cliques (BTND):
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/master/src/JunctionTreeTypes.jl#L259-L262

Perhaps parametric is currently storing messages in either tree
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/99a0620fc2fdc8eb98c28667697188f47589af87/src/JunctionTreeTypes.jl#L23
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/master/src/JunctionTreeTypes.jl#L203-L204

Affie:
> Perhaps parametric is currently storing messages in either tree

see comment about making a distinction between channels and stored messages: https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-609015360

dehann:
[...]

PR #857, opened 2020-08-24
Parametric "take! and store" message again

### LINKED ISSUE #740: restore parametric covariance msgs
broken during #459 work in #739 

2e977a1 

xref: https://github.com/JuliaRobotics/IncrementalInference.jl/commit/2e977a1e4073d76ab8c94a43fd6f8098782a19e8#r39887387

Affie:
Close #740

### DIFF
--- a/src/JunctionTree.jl
+++ b/src/JunctionTree.jl
@@ -756,6 +756,16 @@ function initTreeMessageChannels!(tree::MetaBayesTree)
   return nothing
 end
 
+function refTreeMessageChannelsFromBTND!(tree::BayesTree)
+  for inci in tree.bt.inclist
+    for e in inci
+        downMsg = e.target.data.dwnMsgChannel
+        upMsg = e.target.data.upMsgChannel
+        push!(tree.messages, e.index=>(upMsg=upMsg,downMsg=downMsg))
+    end
+  end
+  return nothing
+end
 
 function appendUseFcts!(usefcts,
                         lblid::Symbol,
--- a/src/CliqStateMachine_Parametric.jl
+++ b/src/CliqStateMachine_Parametric.jl
@@ -92,25 +92,31 @@ function waitForUp_ParametricStateMachine(csmc::CliqStateMachineContainer)
   # csmc.drawtree ? drawTree(csmc.tree, show=false, filepath=joinpath(getSolverParams(csmc.dfg).logpath,"bt.pdf")) : nothing
 
   childrenOk = true
-  beliefMessages = Dict{Int,LikelihoodMessage}()
+  # JT empty upRx buffer to save messages, TODO It may be ok not to empty 
+  beliefMessages = empty!(messages(csmc.cliq).upRx)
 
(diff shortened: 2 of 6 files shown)


Affie:
Let's discuss this in #855

Affie:
The main difference here is:
- `fetch` vs `take!` #855
- storing channels in `edges` vs `nodes` (easily changed)

Affie:
The point of this pr is to get all the current parametric features working again and then change from there.

Affie:
I think I understand the fundamental difference that still remains between parametric and mm CSMs.
Issues #855 and #675 goes hand in hand. 
- fetch is easier with channels in nodes  
- take! is easier with channels in edges 

I don't think we are going to agree on what strategy to follow, so why not have both for a while. After the down part of #459 is done, it can look close enough with only function names different for fetch vs put!.

dehann:
How far is this work (in the off chance this PR does get merged) away from changing PCSM onto a `fetch` model, so that it matches CSM? https://github.com/JuliaRobotics/IncrementalInference.jl/issues/855#issuecomment-679921053

Affie:
> ... changing PCSM onto a fetch model ...

Maybe I'm missing something, but I don't see that as a good option. To do that I need to add conditions to wait on. Why do that if you can use the channels directly?

src/JunctionTree.jl
```diff
@@ -756,6 +756,16 @@ function initTreeMessageChannels!(tree::MetaBayesTree)
...
+function refTreeMessageChannelsFromBTND!(tree::BayesTree)
+  for inci in tree.bt.inclist
+    for e in inci
+        downMsg = e.target.data.dwnMsgChannel
+        upMsg = e.target.data.upMsgChannel
+        push!(tree.messages, e.index=>(upMsg=upMsg,downMsg=downMsg))
+    end
+  end
+  return nothing
+end
```
Affie, line 768: Based on https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-679954452
A container that keeps all channels in one place is also handy for debugging and monitoring. we can also send termination messages to cleanly exit as with PCSM

dehann:
reluctantly removing the blocker, with condition that we work to get both PCSM and CSM to use the same model from your examples #675 -- CSM will soon only be model 1. and sounds like PCSM is model 2.

For completeness, CSM used to be somewhere between model 1 and 3, and after #459 will be only model 1.

Affie:
> reluctantly removing the blocker, with condition that we work to get both PCSM and CSM to use the same model from your examples #675 -- CSM will soon only be model 1. and sounds like PCSM is model 2.

Yes,  That is correct PCSM is 2.

Condition is fine, so I'm going to merge this as is.

Just to recap on our conversation:

The difference between the 2 designs is `fetch` vs `take!`
- fetch is likely to require conditions as well (as in current CSM)
- take! will likely not require additional conditions, especially if an n-stroke solve works correctly

Issue #852, opened 2020-08-22
suppress verboseness of old dev warnings


dehann:
```julia
┌ Warning: Factor x3x4f1 will be an orphan in the destination graph, and therefore not added.
└ @ DistributedFactorGraphs ~/.julia/packages/DistributedFactorGraphs/hftpL/src/services/AbstractDFG.jl:955

┌ Warning: PPE 'default' does not exist, adding
└ @ DistributedFactorGraphs ~/.julia/packages/DistributedFactorGraphs/hftpL/src/services/DFGVariable.jl:676
```

I'd recommend adding a verbose::Bool=true default and then fixing IIF to suppress those warnings.

Affie:
Also, it's not old dev warnings, its what we decided for the API:
On `update` verb, if it (noun) does not exist  `add` with waring.

dehann:
root cause is not that easy in this case — it’s a question how to use the newly consolidated DFG subgraph API.  For this use case in IIF, the list of factors is only known after discovering which factors go between a list of variables.  This is basically the copySubgraph functionality.  The decision discussion conclusion we came to (to not allow orphan factors, i forget the two issues where we wrote that down) was the best compromise solution to keep DFGFactor immutable.

So the use case here is to get a subgraph without knowing what the factors are upfront, and that should just be part of the DFG subgraph functionality as it is now (already implemented but has the warning). Suppressing the warning seems like the best solution to me in this case, otherwise there is too much duplication and a regression on the work we just did.

Note, the use-case here to DFG is “please get the subgraph from variable list so that the associated factors can be discovered”.  The performance consideration is “don’t trudge through all factors to find a clean list” of factors for an over-simplified copy subgraph.  DFG API requirement is “copy subgraph from only a list of variables, with option to include factors that have a subset of associated variables”. Think we partially ditched the last optional part “include factors with subset of associated variables” in order to preserve “DFGFactor is immutable”.

Affie:
Ok, so its a case of... 
- It's on by default, so take care if you see it, you might have done something wrong.
- If you know what you are doing, you can suppress the warning. 

xref to DFG issue TBD

Affie:
I'm not keeping this for the comment, thanks for removing it.
This PR is open to remove warnings from `update`

PR #858, opened 2020-08-24
Quiet buildSubraph with orphans

Affie:
part of #852


dehann:
Thanks!  I don’t think the FIXME comment is needed in this case, see discussion in #852

PR #861, opened 2020-08-27
WIP 3 consolidate dwnMsg

dehann:
part of #459

### DIFF
--- a/src/CliqStateMachine.jl
+++ b/src/CliqStateMachine.jl
@@ -476,10 +476,9 @@ DevNotes
 """
 function attemptCliqInitDown_StateMachine(csmc::CliqStateMachineContainer)
   #
+  # TODO consider exit early for root clique rather than avoiding this function
   infocsm(csmc, "8a, needs down message -- attempt down init")
   setCliqDrawColor(csmc.cliq, "gold")
-
-  # TODO consider exit early for root clique rather than avoiding this function
 
   # initialize clique in downward direction
   # not if parent also needs downward init message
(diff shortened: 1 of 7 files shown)


src/CliqStateMachine.jl
```diff
@@ -532,29 +529,28 @@ function attemptCliqInitDown_StateMachine(csmc::CliqStateMachineContainer)
...
   unlockUpStatus!(prnt) # TODO XY ????
   infocsm(csmc, "8a, attemptCliqInitD., unlocked")
 
 
   # FIXME try split CSM here, need replacement for dwinmsgs
   dwnkeys_ = ls(csmc.cliqSubFg, tags=[:DOWNWARD_COMMON;]) # FIXME THIS DOESNT WORK YET!
-  # @assert intersect(dwnkeys, dwnkeys_) == length(dwnkeys) "split dwnkeys_ is not the same, $dwnkeys, and $dwnkeys_"
-
+  # FIXME, why is this breaking the IIF and RoME tests?
+  @assert intersect(dwnkeys, dwnkeys_) == length(dwnkeys) "split dwnkeys_ is not the same, $dwnkeys, and $dwnkeys_"
```
Affie, line 542: I tried a bit, but don't follow completely. 

- the assert `intersect(dwnkeys, dwnkeys_) == length(dwnkeys)` is a set and a length
- `DOWNWARD_COMMON` is added to factors, so I thought the idea was to get the variables attached to the factors with that tag.

Here is what I tried, it works for a few, but there are cases with no DOWNWARD_COMMON tag at all.

```julia
  # FIXME try split CSM here, need replacement for dwinmsgs
  @info "DOWNWADS_COMMON $(lsf(csmc.cliqSubFg, tags=[:DOWNWARD_COMMON;])) --- $dwnkeys"
  dwnkeys_ = Symbol[]
  union!(dwnkeys_, getNeighbors.(csmc.cliqSubFg, lsf(csmc.cliqSubFg, tags=[:DOWNWARD_COMMON;]))...)
   # FIXME THIS DOESNT WORK YET!
  # FIXME, why is this breaking the IIF and RoME tests?
  @assert issetequal(dwnkeys, dwnkeys_) "split dwnkeys_ is not the same, $dwnkeys, and $dwnkeys_"
```

Affie: Ahh, I was close. will leave down messages to you.

Issue #865, opened 2020-08-28
add test coverage for DRT


dehann:
https://codecov.io/gh/JuliaRobotics/IncrementalInference.jl/src/master/src/TetherUtils.jl

as per:
https://juliarobotics.org/Caesar.jl/latest/examples/deadreckontether/

import to establish convention of:
```julia
fg = initfg()

addVariable!(fg, :x0, Pose2, tags=[:POSE;], solvable=0) # or Pose3
# ... add factors etc
# set that graph fragment solvable

addVariable!(fg, :drt_0, Pose2, tags=[:POSE; :DRT], solvable=0)
drec1 = MutablePose2Pose2Gaussian(...) # or InertialPose3
addFactor!(fg, [:x0; :drt_0], drec1, solvable=0, graphinit=false)

# no find high frequency solution with:
accumulateDiscreteLocalFrame!
accumulateFactorMeans

# and drop a new pose variable with
duplicateToStandardFactorVariable
```

PR #873, opened 2020-09-01
Test dead reckoning tether based on RoME test

Affie:
part of #865


Affie:
> cont2disc

Do you have an example or something on this? I don't know how to test it.

dehann (approved):
thanks, will add cont2disc when i get an example.  Its the same as c2d in matlab, but this one does covariance in the proper way too.  Should be the easiest and best way to build Kalman Filters etc if needed.

PR #884, opened 2020-09-08
fast-forward csm monitor task work

### LINKED ISSUE #459: Consolidate belief messages and message channels

dehann:
Hi @Affie please merge as soon you can.  Just a heads up that more changes are coming to CSM soon as part of the push to close #459 .


dehann:
I would actually also ask to hold off on task monitoring or any other new features in PCSM.  I'm really nervous about old code being brought in by git mistake or otherwise and getting a regression in CSM which is going to be hard to resolve.  I'd actually request that any new work (like task monitoring) be brought in through CSM first and not via PCSM.  New features coming in via PCSM are just going to make the consolidation with CSM more difficult.  PCSM has to follow CSM from here on out please.  CSM should be the testing ground for new ideas, not PCSM.  In October we are going to ditch the name PCSM entirely, and should only rely on a config to have either default or parametric states run from a single common CSM.

Affie:
> I would actually also ask to hold off on task monitoring or any other new features in PCSM

I'm not sure I'm following, the task monitor has been in PCSM for a while. This branch was experimenting with adding it to CSM. It can work with throwing exceptions like this, but to exit cleanly would be better. That can only happen after #459

dehann:
Alright, that sounds good thanks.  Sorry I have not been stopping to look at any code or PRs.  Just focussing on getting #459 done.  Would be great to have the error termination code in CSM once 459 is done.

PR #885, opened 2020-09-09
WIP 6 refactor csm in prep dwnMsg consolidation

dehann:


src/CliqStateMachine.jl
```diff
@@ -820,9 +820,85 @@ end
...
+Redirect CSM traffic in various directions 
+
+Notes
+- State machine function nr.4b
+- Was refactored during #459 dwnMsg effort.
+
+DevNotes
+- Consolidate with 7?
+"""
+function trafficRedirectConsolidate459_StateMachine(csmc::CliqStateMachineContainer)
```
Affie, line 835: I like this. I like the idea of branching in as few states as possible.

Issue #674, opened 2020-04-04
CSM message passing push or pull model


dehann:
## Problem

Assuming inter-clique data is stored in CSMC/BTND ~~(although does not change too much if stored on the tree edges)~~, should tree message channels be stored in the sending or receiving CSMC/BTND?  E.g., should upward message channels be stored in child or parent CSMC/BTND?

### Related:
- #459 (first nonparametric/parametric consolidation effort)
- #673 (super solving)
- #675 (store in csmc or tree edges)

## Considerations

- Historically tree initialization requires a dictionary of up messages message channels are stored in the parent.
- Nonparametric currently uses "push" model with channels stored in BTND.
- Parametric currently stores messages in a separate channel on tree and linked by edges.
    - JT Edit: This is not correct, the channels are only "wires" used for communication.
    - Currently messages are stored in the CSMS as I wanted to run both mm and parametric and compare results. The 2 options I can see (I'm ok with both) is store it in CSMS or in a dictionary in the clique under solvekey as in the DFG implementation.

## Pros and Cons

| push (push msg to receiver) | pull (keep msg in local) |
|--------|------|
| :heavy_check_mark: solve is self contained | :heavy_multiplication_x: requires message fetch |
| :heavy_multiplication_x: parent needs dict of msgs | :heavy_check_mark: all msgs containers are singular |

dehann:
Current CSM code looks like `cliqSolverData().upInitMsgs` needs a dictionary from children upward messages
- Looks like push model -- i.e. stored in parent.

Affie:
Parametric uses `Channel`s to move data between cliques, therefore its a push/pull (`put!`/`take!`) model to/from the channel. 
eg. For up, children pushes to up channel and parents pulls, blocking until all messages are received. 
Currently no parent/child cliques are needed, only the channels. Therefore, the storing of channels in the tree and not CSMC. 

This method definitely made it easier to develop parametric, however it still doesn’t do initialization.

dehann:
> therefore its a push/pull (put!/take!) 

Right, so push/pull has double meaning and I meant something else:  where is the physical location of the message channels?  Either on the edges or in parent/child csmc or btnd, or separately in the tree?  This is part of #675 but still slightly different, and only comes into play of we choose to store the channels in either BTND or CSMC -- I think this is the same reason why you liked storing channels on the tree (linked from edges).

[...]

PR #892, opened 2020-09-10
transfering CSM 8a into parent

dehann:
Mental model for what steps might be for untangling "CSM 8a pull model (#674)" which is required to resolve "#459 dwnMsg consolidation":
- redirect 4b needdownmsg to 7 instead (replacing 4c direction)
- absorb 4c into 7
- redirect 7 away from 7c, but to 8j instead
- add path from new 7 to 8a
- Redirect 8a away from 8j, but perhaps to 7b instead
- Finally fix 8a to pull model which should now be running in parent csm instead.


dehann:
finalizes pull model on #674

PR #901, opened 2020-09-16
Monitor CSM tasks and send InterruptException + tests

Affie:


dehann:
HI @Affie, thanks -- i have not had time to look at this.  #459 is all consuming at this time.  Also note that I have a massive pile of changes to CSM on my local machine and might add a few more conflicts.  All this will go from `twig/3Q20/dwn459/fix674` into `maint/20Q3/fix459` and then into master.  The changes will require IIF v0.16.0 so lets delay this until then to merge?

dehann:
PS, also see the new FOR DEBUG USAGE ONLY `solveTree!(fg, timeout=50)` feature in #892

Affie:
Ok, no problem. I’ve been using it and found it helpful for development mistakes. So cleaned it up in case it can help. 
I saw the timeout feature and was thinking of updating it to a per state (function) value. 
Something like get(per_fnc_timeout, fnc, nothing) so you can have longer timeout for the upsolve for instance. 
But, in the long run a watchdog timer on a heart beat might work better. For stalled cliques, the task monitor seems to work well for errors.

dehann:
cool yeah, as before --  459 is almost done, then consolidate parametric into CSM (dropping PCSM entirely), and then only look at new features.  Fixing the poor maintainability of the code is most critical thing at this point, so lets fix that first and then start adding new stuff to the consolidated code base after.  I'm busy debugging IIF test failures on new CSM stuff here and should then be able to start merging recent dwnMsg 459 consolidation work.

dehann (approved):
Hi @Affie , this looks great thanks!  If you would not mind clearing the conflicts with the new code in master and merge as soon as possible please.  Think we should target this for v0.16.0 and if we can actually tag IIF as such around Monday or Tuesday.

PR #906, opened 2020-09-17
update MultiHypo2Door

Affie:
TODO 
- [ ] `doCliqInferenceUp!` replacement cc @dehann
  - SEe wip #913

### DIFF
--- a/src/SolveTree.jl
+++ b/src/SolveTree.jl
@@ -606,17 +606,18 @@ function treeProductUp(fg::AbstractDFG,
 
   # get all the incoming (upward) messages from the tree cliques
   # convert incoming messages to Int indexed format (semi-legacy format)
-  upmsgssym = LikelihoodMessage[]
-  for cl in childCliqs(tree, cliq)
-    msgdict = getUpMsgs(cl)
-    dict = Dict{Symbol, TreeBelief}()
-    for (dsy, btd) in msgdict.belief
-      vari = getVariable(fg, dsy)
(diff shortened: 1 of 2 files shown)


src/SolveTree.jl
```diff
@@ -606,17 +606,18 @@ function treeProductUp(fg::AbstractDFG,
...
+  #   msgdict = getUpMsgs(cl)
+  #   dict = Dict{Symbol, TreeBelief}()
+  #   for (dsy, btd) in msgdict.belief
+  #     vari = getVariable(fg, dsy)
+  #     # manis = getSofttype(vari).manifolds
+  #     dict[dsy] = TreeBelief(btd.val, btd.bw, btd.inferdim, getSofttype(vari))
+  #   end
+  #   push!( upmsgssym, LikelihoodMessage(beliefDict=dict) )
+  # end
+  upmsgssym = getMsgsUpChildren(tree, cliq, TreeBelief)
```
Affie, line 620: @dehann please confirm

dehann: Hi @Affie , yes that looks good.  

Just in case you were not aware, `treeProductUp` was added to help with debugging on the tree and has not yet been folded into the main solution plumbing yet.  It is unclear if `treeProductUp` will actually work for regular solving, because it fetches new up messages and does a single convolve-product step for the particular variable.  The main workhorse in the upward solve code at the moment is `upGibbsCliqueDensity`.  We can figure out how to best consolidate these two use cases at some point.

PR #907, opened 2020-09-17
update MultiHypo3Door

Affie:


examples/MultiHypo3Door.jl
```diff
@@ -305,10 +302,10 @@ ensureAllInitialized!(fg)
 
 
 tree = wipeBuildNewTree!(fg, drawpdf=true, show=true)
-cliqorder = getCliqOrderUpSolve(tree)
```
Affie, line 308: @dehann, do you perhaps know the replacement for deprecated `getCliqOrderUpSolve`

src/FactorGraph.jl
```diff
@@ -880,7 +880,7 @@ function initManual!(dfg::AbstractDFG, sym::Symbol, pX::BallTreeDensity)::Nothin
 end
 function initManual!(dfg::AbstractDFG, sym::Symbol, usefcts::Vector{Symbol})::Nothing
   @info "initManual! $sym"
-  pts = predictbelief(dfg, sym, usefcts)
+  pts = predictbelief(dfg, sym, usefcts)[1]
```
Affie, line 883: This looks like a bug, found with MultiHypo3Door example

dehann: oh thats weird, I had been using that quite a bit at some point... Ah think i see what happened:
```julia
return pGM, sum(inferdim)
```

The newer code needs both the `points` and the `inferdim`.  Just make sure that the use cases inside and below `initManual` do not also require `inferdim`?  Perhaps the name `pts` is misleading in the newer version

examples/MultiHypo3Door.jl
dehann: `getEliminationOrder`

dehann:
Hi @Affie , not sure what you wanted to do here still, perhaps you could help finish this when ready and PR to master as needed?  Just want to close out loose ends if possible please.

src/FactorGraph.jl
Affie: Doesn't look like it, its only `Xpre = AMP.manikde!(pts, getSofttype(vert) |> getManifolds )`

PR #912, opened 2020-09-23
upgrade mixture prior, needed downstream

dehann:


dehann:
xref JuliaRobotics/DistributedFactorGraphs.jl#668

PR #929, opened 2020-09-29
maint and Common CSM, fix MixtureRelative

dehann:
Hi @Affie , note I moved some CSM functions into a common_consolidate file.  Most of these functions should be agnostic to the solver being used, and likely a few more in CSM.jl that could move over too.  Just wanted to move some over to get the story started.  Check if you can add these to the `take!` model system perhaps please.  If both solvers enter and exit CSM the same, and have a few functions in common, then that really helps limit the scope of differences between `fetch` and `take` models.

ref #855


Affie:
I tried to find the common points and so far it looks like the best places might be:

  - `blockUntilChildrenHaveStatus_StateMachine` <==> `waitForUp...`
  - `waitChangeOnParentCondition_StateMachine` <==> `waitForDown...`


`blockUntilChildrenHaveStatus_StateMachine` fits well and the 2 are (can be) roughly equivalent up to there.

dehann:
Hi Johan, that sounds right.  Probably also 
- `slowIfChildrenNotUpsolved` <==> `waitForUp`

PR #932, opened 2020-09-30
Consolidate solveTree! and solveTreeParametric!

Affie:


src/CliqStateMachine_Parametric.jl
```diff
@@ -1,4 +1,32 @@
-
+"""
+    $SIGNATURES
+Helper function to log a message at a specific level to a clique identified by `csm_i` where i = cliq.index
+"""
+function logCSM(csmc, msg::String; loglevel::Logging.LogLevel=Logging.Debug, maxlog=nothing, kwargs...)
```
dehann, line 5: Hi, probably best to consolidate with the existing function before merging, but this looks good.  The most important thing for me on `infocsm` is that it gets attached to a different **`csmc.logger`** for each CSM.  This produces the independent `/tmp/caesar/DATETIME/logs/cliq%d/log` files.  I remember you mentioning that you want to split the logging information after recording one file.  Just make sure you check on existing logging code before building two separate logging stacks.  It's okay to use the best/consolidated approach, but lets make sure there is only one of everything.    I'd also ask to please move this function out of the Parametric file into CSMCommon_Consolidate.jl please.
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/31a0d1718a64203da4194cf5ff70421e24caa05f/src/CSMCommon_Consolidate.jl#L11-L25

src/CliqStateMachine_Parametric.jl
```diff
@@ -306,10 +330,18 @@ function waitForDown_ParametricStateMachine(csmc::CliqStateMachineContainer)
       end
       @error "Par-4, $(csmc.cliq.index): Exit with error state"
       return IncrementalInference.exitStateMachine
+    
+    elseif beliefMsg.status == :DOWNSOLVED
```
dehann, line 334: just check CSM uses lower case `:downsolved` -- I had a multihour debug issue on upper/lower case status the other day.  Qe can go back to enums whenever is best.

dehann (approved):
super cool!

src/CliqStateMachine_Parametric.jl
Affie: Hi, yes. This started out as something I experimented with and then started using out of convenience. I only use it for debugging currently and do not want to replace the current `infocsm`. It would be a big change in the logging design and not something worth doing unless we are 100% sure it would be a big improvement.

src/CliqStateMachine_Parametric.jl
Affie: Thanks, looks like I'm still on uppercase for up and down.

src/CliqStateMachine_Parametric.jl
Affie: In effect, it's just a convenient @debug that groups the message by clique. I prefer the REPL over files.

PR #945, opened 2020-10-05
_dbgCSMSaveSubFG helper function used to save debug subFGs

Affie:


src/Deprecated.jl
```diff
@@ -1058,7 +1058,36 @@ function updateTreeCliquesAsMarginalizedFromVars!(fgl::AbstractDFG, tree::Abstra
...
+Internal helper function to save a dfg object to LogPath during clique state machine operations.
+
+Notes
+- will only save dfg object if `opts.dbg=true`
+
+Related
 
+saveDFG, loadDFG!, loadDFG
+"""
+function _dbgSaveDFG(dfg::AbstractDFG,
```
Affie, line 1076: # NOTE I only saw this function after I replace all the functions with _dbgCSMSaveSubFG
# I consolidated the 2 and think this one can therefore be deprecated. Unless you use it 
# separate from a CSMC

dehann: hi, thanks perhaps want to just add a '@deprecate' line and remove the outdated function?

Issue #910, opened 2020-09-19
Consolidate btnd.solvableDims into LikelihoodMessage


dehann:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/c7b3c3d91ae586c63d68f51b15e839ef7422cb7b/src/JunctionTreeTypes.jl#L347

dehann:
Okay, so logging how to solve this one.  Each sibling has a possible solvableDim per variable in the clique.  

The rule of thumb here is that if the solve were to occur one sibling at a time, then the last sibling would have the most information available to resolve its estimates;  simultaneously, the first sibling (and intermediate ones) may or may not have the required information from priors yet propagated to that part of the factor graph (from any prior and through measurement functions).

Given the current level of solved-ness of each variable (during init) some cliques are further away from being solved than others.  This information needs to be communicated to each sibling somehow, so that the best order of precedence in various situations can be resolved.  This communication should happen according to tree construction and messages passed along the edges only.

dehann:
The old update call is:
```julia
sdims = getCliqVariableMoreInitDims(csmc.cliqSubFg, csmc.cliq)
updateCliqSolvableDims!(csmc.cliq, sdims, csmc.logger)
```

Affie:
I'm still not completely following with `solvableDims`
Say we work with hex but with priors on x3 and x5 on this tree:
![image](https://user-images.githubusercontent.com/6612981/95227785-d1ed2780-07fe-11eb-9728-d91b3a8477a4.png)

5 and 6 can both do an upinit, why can't they go at the same time?
They have x4 in common so is the information not best mixed in the parent (2)?

dehann:
still wip to get the new code to act the same.

dehann:
Hi @Affie , check out the previous discussion in #344

dehann:
not going to make this 

> at the same

and parking sibling cascade order in #954 (old fetch model).

PR #948, opened 2020-10-06
take! mm tree init poc

Affie:
- As separate functions for now, but can be integrated with existing take!. It is also reusing most of `fetch`'s important bits.
- It can only work with `useMsgLikelihoods`


Affie:
Octo tree init

dehann:
This together with post #910 `solvableDims` should be a strong combo.  That keeps the door open for the cascade model and removes the last of the tangled mess from old CSM.

I think we should merge this PR into master now before v0.16 and just make it part of the v0.17.0 consolidation process.  I don't think we should leave this branch hanging on v0.0.x.  Johan, any objections?

Affie:
I was also thinking we could merge. That's why its separate files in experimental. 
If it is merged I can consolidate with parametric quite easily. 
I will update milestone

PR #949, opened 2020-10-06
Bug fix: save :initialized and use graphinit kwarg on _lineStep

Affie:


src/SubGraphFunctions.jl
```diff
@@ -128,7 +128,7 @@ function transferUpdateSubGraph!(dest::AbstractDFG,
   # transfer specific fields into dest from src
   for var in (x->getVariable(src, x)).(syms)
     # copy not required since a broadcast is used internally
-    updateVariableSolverData!(dest, var, solveKey, false, [:val; :bw; :inferdim; :solvedCount]; warn_if_absent=false)
+    updateVariableSolverData!(dest, var, solveKey, false, [:val; :bw; :inferdim; :solvedCount; :initialized]; warn_if_absent=false)
```
Affie, line 131: with tree init, :initialized is set in subfg but not updated in full fg. There might be other flags also needed.

src/CanonicalGraphExamples.jl
```diff
@@ -145,7 +145,7 @@ function generateCanonicalFG_lineStep(  lineLength::Int;
         if mod(i,poseEvery) == 0
             push!(x, i)
             addVariable!(fg, Symbol("x",i), vtype) #, autoinit = graphinit)
-            (i in posePriorsAt) && addFactor!(fg, [Symbol("x",i)], Prior(xNoise(i, σ_pose_prior)))
+            (i in posePriorsAt) && addFactor!(fg, [Symbol("x",i)], Prior(xNoise(i, σ_pose_prior)), graphinit=graphinit)
```
Affie, line 148: I fixed this before, but I think it got lost in a merge somehow.

PR #952, opened 2020-10-07
Use empty FG in addLikelihoodsDifferential!

### LINKED ISSUE #950: Manual init cliqSubFg in addLikelihoodsDifferential!
Manual initialization in `addLikelihoodsDifferential!` can overwrite the clique's own initialization and can cause (if its wrong) the clique to take longer to converge. 
A possible solution is to use a separate empty fg.

https://github.com/JuliaRobotics/IncrementalInference.jl/blob/34eb45287466fce16cb3f89bbad7327f2bef8789/src/TreeMessageUtils.jl#L154-L165

Affie:
close  #950

### DIFF
--- a/src/TreeMessageUtils.jl
+++ b/src/TreeMessageUtils.jl
@@ -151,7 +151,7 @@ Notes
 DevNotes
 - Initial version which only works for Pose2 and Point2 at this stage.
 """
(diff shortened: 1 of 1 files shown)


dehann (approved):
Looks good, thanks!

Issue #754, opened 2020-06-20
CSM init endless cycle MWE fail case 


dehann:
Found a easy fail case on CSM, relates to existing test:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/03d231d63fad0f7b2ea897c328a657753ca8d7c2/test/testBasicTreeInit.jl#L9

### MWE

Clique 5 and 6 solves cycle endlessly if the following clique solve sequence is used:

```julia
fg = generateCanonicalFG_CaesarRing1D(graphinit=false)
getSolverParams(fg).graphinit = false
getSolverParams(fg).limititers = 50
TS = Vector{Task}(undef,6)
TS[5] = solveCliq!(fg, tree, :x5, async=true)  # , recordcliq=true
TS[6] = solveCliq!(fg, tree, :x3, async=true)  # , recordcliq=true
```

dehann:
also added as a recent new test in IIF, testBasicTreeInit.jl

Affie:
Another basic endless cycle fail case (after #459)
```julia
using IncrementalInference
fg = generateCanonicalFG_lineStep(5; 
                                  poseEvery=1, 
                                  landmarkEvery=5, 
                                  posePriorsAt=[0,2], 
                                  sightDistance=4,
                                  solverParams=SolverParams(algorithms=[:default, :parametric]))
                                  
getSolverParams(fg).graphinit = false
getSolverParams(fg).treeinit = true
smtasks = Task[]
tree, smt, hist = solveTree!(fg; smtasks, verbose=true, timeout=50, recordcliqs=ls(fg));
```

Affie:
![image](https://user-images.githubusercontent.com/6612981/94006499-37252f80-fda0-11ea-8d72-b4272000f8ec.png)

dehann:
I'm going to debug a little by adding:
```julia
fid = open(joinLogPath(fg,"csmVerbose.log", "w")
solveTree!(...;  verbosefid=fid, verbose=true, ...)

# after finished
close(fid)


open(joinLogPath(fg, "csmLogicalReconstructMax.log"),"w") do io
  IIF.reconstructCSMHistoryLogical(getLogPath(fg), fid=io)
end
```

dehann:
Here it is:
[csmLogicalReconstructMax.log](https://github.com/JuliaRobotics/IncrementalInference.jl/files/5280411/csmLogicalReconstructMax.log)

[...]

PR #953, opened 2020-10-07
WIP to fix #910

### LINKED ISSUE #910: Consolidate btnd.solvableDims into LikelihoodMessage

dehann:
Moving some code around to help with understanding sibling down init ordering.  Relates to #910, #754


dehann:
FYI @Affie , just check what i'm doing here -- more coming, I'm going to move utility content for "sibling down init priority ordering" into `CSM_fix910.jl`.  There are some bad hacks in there that need to sorted out.  Think these set of fixes together should be the resolution to the tangled CSM state.

I will add to these new `8o.i - 8o.iv` where the bad hacks a of non-channel information sharing between the CSMs.

One thing that is clear from this PR is that there is strong need for repeated fetch/take of both up and especially down messages in CSM API.  So however #855 works out, this repeated fetch/take of the same message looks to be a real requirement.

Affie:
> One thing that is clear from this PR is that there is strong need for repeated fetch/take of both up and especially down messages in CSM API. So however #855 works out, this repeated fetch/take of the same message looks to be a real requirement.

I just see the fetch in this case as a buffer of previous values. I would guess that it would be bad if the parent got new information and updated while in the middle of trying to figure the order out.
The way I see it fetch and take! both have a put/wait/take! function and a buffer function and both are needed (even if it just makes thing cleaner and easier).

Issue #954, opened 2020-10-07
Parent CSM should resolve sibling init order (if partials or similar)


dehann:
Current CSM states `8o.ii`, `8o.iii`, and `8o.iv` require siblings directly request each other's status.  That is not according to tree edges, and can therefore likely only occur in the parent (if following structure).

Stems from
- #754, 
- #910

Also see pull model,
- #674 

cc @Affie

dehann:
see xref #955 for relevant code in `src/CSM_SiblDwnInit_fetch.jl`

dehann:
Previously (old fetch CSM model) in cascade tree-init, when siblings cannot resolve their initialization there is a problem of figuring out which sibling should be initialized first.  The old fetch code required each clique to resolve the init priority of all its siblings (basically sharing data not along edges of the tree -- really bad).  

If cascading and sibling order is required again, these old functions should be performed by the common parent instead and then direct which child to solve first.

dehann:
I believe this issue was closed with new CSM and X stroke.  Cascading might come back later.

PR #958, opened 2020-10-08
Only downsolve when all children finished upsolve

### LINKED ISSUE #924: parent starting down solve too early bug
When solving `Caesar1D` or `Hex` with tree init, clique 1 starts downsolve before clique 2 completes it's upsolve

Affie:
close #924

Tests:
- Passed RoME tests locally
- Passed with #957 also locally


dehann (approved):
thanks, yes lets go with this -- just good to note the down init sequencing between siblings is now 'wrong'.  Although work on #910 and #954 will directly address that issue anyway, so in terms of progress lets merge this and keep moving.  Tree init is still off by default and experimental so does not directly affect active users.

dehann:
ah dang, this recreates #754 issue:

```
fg = generateCanonicalFG_Circle(8;graphinit=false)
getSolverParams(fg).graphinit = false
getSolverParams(fg).treeinit = true
# getSolverParams(fg).drawtree = true
# getSolverParams(fg).showtree = true

solveTree!(fg);
```

Affie:
dang, I thought IIF and RoME got all the cases that were failing. Adding test in #959

PR #977, opened 2020-10-17
Consolidate parametric and x

Affie:


src/experimental/CliqStateMachine_X-stroke.jl
```diff
@@ -310,23 +298,19 @@ function waitForDown_X_StateMachine(csmc::CliqStateMachineContainer)
     logCSM(csmc, "$(csmc.cliq.index): Belief message received with status $(beliefMsg.status)")
 
     logCSM(csmc, "X-4 down msg on $(keys(beliefMsg.belief))"; beliefMsg=beliefMsg)
-    # save DOWNSOLVED incoming message for use and debugging
+    # save down incoming message for use and debugging
     messages(csmc.cliq).downRx = beliefMsg
```
dehann, line 302: does Rx mean "push" model?  I have only "pull" model in mind #674 and:
https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-679938712

Affie: It is the received message from the channel. In this case, channels are zero length and therefore not buffered. So its the buffer used in place of fetch but always stored locally in the clique that uses it and never access from outside in a pull or push way. Currently, there are no "push/pull" in the same sense as 674 as the parent and child cliques are not stored in the CSMC. The "pull" storage model for channels https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-679938712 is less flexible as you cannot only send a message to one child branch while blocking on the other see #675

Issue #925, opened 2020-09-27
Internal clique variable solver order using `cliqSubFg` not old `btnd.potentials`


dehann:
Post #913 and #923 , the old way of deciding which variables to `iterVar` or solve `direct` inside a clique during upsolve is likely no longer correct.  Should use `csmc.cliqSubFg` directly to figure out what is the best order, and who to skip + direct solve.

dehann:
will influence BTND fields
```julia
  cliqAssocMat::Array{Bool,2}
  cliqMsgMat::Array{Bool,2}
  directvarIDs::Vector{Symbol}
  directFrtlMsgIDs::Vector{Symbol}
  msgskipIDs::Vector{Symbol}
  itervarIDs::Vector{Symbol}
  directPriorMsgIDs::Vector{Symbol}
```

PR #979, opened 2020-10-19
getCliqVarInitOrderUp use subgraph

### LINKED ISSUE #978: Bug in getCliqVarInitOrderUp
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/870e518fc7e808b911bcd50682a2f786a793a80b/src/TreeBasedInitialization.jl#L25-L65

`upmsgs` is a `Dict{Int, LikelihoodMessage}` which make `upmsgids` a `Vector{Int}`
while `allids` is a `Vector{Symbol}`

Affie:
close #978
xref #925

### DIFF
--- a/src/TreeBasedInitialization.jl
+++ b/src/TreeBasedInitialization.jl
@@ -12,7 +12,6 @@ function isCliqUpSolved(cliq::TreeClique)::Bool
 end
 
 
-
 """
     $SIGNATURES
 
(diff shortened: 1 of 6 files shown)


src/TreeBasedInitialization.jl
```diff
@@ -12,59 +12,38 @@ function isCliqUpSolved(cliq::TreeClique)::Bool
...
+- sorts id (label) for increasing number of connected factors using the clique subfg with messages already included.
 """
-function getCliqVarInitOrderUp(tree::BayesTree, cliq::TreeClique)
+function getCliqVarInitOrderUp(subfg::AbstractDFG)
   # rules to explore dimension from one to the other?
 
   # get all variable ids and number of associated factors
-  allids = getCliqAllVarIds(cliq)
-  nfcts = getCliqNumAssocFactorsPerVar(cliq)
+  B, varLabels, facLabels = getBiadjacencyMatrix(subfg)
```
dehann, line 28: yep ,this is good -- later i want to use AMD, RCS, or CCOLAMD to work inside a clique just like over the whole graph but that is an improvement.  This is good for now.

For reference:
#980

dehann (approved):
This look good to me, thanks for doing!

PR #1000, opened 2020-10-25
Mixture isPrior

### LINKED ISSUE #998: Bug - Mixture Prior is not an AbstractPrior
```julia
p = Mixture(Prior, [Normal(1,1),Normal(1,1)], [0.5, 0.5] )
isPrior(p) == fasle
p isa AbstractPrior == false
```

This makes functions such as `lsfPriors(subfg)` not find it.

Affie:
needs DFG #679 to be merged and will be tagged as v0.10.7
close #998


dehann (approved):
yeah that looks right, function returns bool.  Mixture can be either prior or relative and either Toots or Minimize.  Looks like i missed one utility here -- thanks

dehann:
*Roots

Issue #567, opened 2020-01-30
CSM to downsolve only from prior upsolved state


dehann:
should be able to call solveTree! to do only a downsolve:
```julia
getSolverParams(fg).upsolve = false
getSolverParams(fg).downsolve = true
```

The upsolve only case is already working.

dehann:
Think this is working and we can test and close?

dehann:
I'll do some better documentation on this

Affie:
We still need the down solve only case.

Affie:
PR #1002 provides the basic structure and is the start of downsolve only.

PR #1002, opened 2020-10-26
Basic skip upsolve state

Affie:
xref #567

### DIFF
--- a/src/CliqueStateMachine.jl
+++ b/src/CliqueStateMachine.jl
@@ -355,8 +355,24 @@ function tryDownSolveOnly_StateMachine(csmc::CliqStateMachineContainer)
 
   logCSM(csmc, "Skipping upsolve clique $(csmc.cliqKey)"; loglevel=Logging.Warn, st=getCliqueStatus(csmc.cliq))
   if getCliqueStatus(csmc.cliq) == NULL 
-    logCSM(csmc, "Clique $(csmc.cliqKey) status NULL, trying as UPSOLVED"; loglevel=Logging.Warn)
-    setCliqueStatus!(csmc.cliq, UPSOLVED)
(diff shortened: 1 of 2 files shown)


src/CliqueStateMachine.jl
```diff
@@ -339,6 +344,24 @@ function solveUp_StateMachine(csmc::CliqStateMachineContainer)
...
+Notes
+- State machine function nr. #XXX
+"""
+function tryDownSolveOnly_StateMachine(csmc::CliqStateMachineContainer)
+  logCSM(csmc, "tryDownSolveOnly_StateMachine clique $(csmc.cliqKey) status $(getCliqueStatus(csmc.cliq))")
+
+  logCSM(csmc, "Skipping upsolve clique $(csmc.cliqKey)"; loglevel=Logging.Warn, st=getCliqueStatus(csmc.cliq))
+  if getCliqueStatus(csmc.cliq) == NULL 
+    logCSM(csmc, "Clique $(csmc.cliqKey) status NULL, trying as UPSOLVED"; loglevel=Logging.Warn)
+    setCliqueStatus!(csmc.cliq, UPSOLVED)
```
dehann, line 359: Not sure skipping upsolve should set the status UPSOLVED, especially coming from NULL.  Trying to think what we can do better here...  Perhaps RECYCLED or SKIPPED is better?

Affie: yes, I had the same worry, that's why the warning. I think you are correct with RECYCLED. Also, it needs some checks to see if it can be recycled.

Affie: @dehann, see changes. I'm going with `solvedCount` > 0 then `UPRECYCLED` else `@error`

PR #1004, opened 2020-10-28
Split down CSM

Affie:
- Split down solve/init from postDown
- Split down solve/init from preDown
- renumber CSM


Affie:
Merging as is, may need improvements for leaves.

PR #1008, opened 2020-10-29
hists as NamedTuple CSMHistoryTuple

Affie:


Affie:
Hi @dehann, this may still have problems with untested code such as some of the animation functions. I guess it may be better to update FSM.jl? 
We can maybe merge now and fix issues as we find them?

dehann:
that sounds good thanks, yes lets updated FSM as we are able to do so

dehann (approved):
looks good thanks

Issue #1010, opened 2020-10-29
Differential factors should be partial if source is only partial


Affie:
For example, if a child clique has a `Pose2Point2Bearing at `l1 | x1,x2` then the message can only be `Pose2Pose2Bearing` ie. partial on θ

![image](https://user-images.githubusercontent.com/6612981/97618303-860d5700-1a27-11eb-95cf-54e5de9d7671.png)

Affie:
This subfg from a range only example
![image](https://user-images.githubusercontent.com/6612981/98226718-11727500-1f5f-11eb-8d2d-0fcaa8d5785c.png)

Results in this factors after deconvolution.
![image](https://user-images.githubusercontent.com/6612981/98226761-23541800-1f5f-11eb-991a-25fcfc3024af.png)

Affie:
For now `useMsgLikelihoods=false` should be used for better performance

Affie:
Example plots above were adapted from the ranges examples from https://github.com/JuliaRobotics/RoME.jl/blob/master/examples/RangesExample.jl

dehann:
thanks, also linking some documentation on this that example for reference: https://juliarobotics.org/Caesar.jl/latest/examples/basic_slamedonut/

dehann:
have not forgotten, just been a bit busy sorry.  Will get to this soon!

dehann:
Hi @Affie, unfortunately I'm not sure how to precisely reconstruct the factor graphs for either of the two examples that you mentioned above.  Could you perhaps post them here please, either the precise factor graphs via script or as simple `saveDFG` objects please (IIF v0.17 or up).

> Needed for parametric tree solve.

#466 is required for the parametric case.

dehann:
please modify and or update a FileDFG of the factor graphs which makes it a bit easier to compare etc.

Affie:
The first one was is just in general and not really a specific example.
I'll put up the code for the examples so you can modify them. I just fiddled with it a lot, so it's currently not the same as the plots.
`:x0` from the above comment is actually `:l100` in the example. I updated it to fix.

Affie:
```julia
using DistributedFactorGraphs
using IncrementalInference
using RoME
using RoMEPlotting
##

GTp = Dict{Symbol, Vector{Float64}}()
GTp[:x0] = [-20.0;0]
GTp[:x1] = [50.0;0]
GTp[:x2] = [100.0;0]
[...]

PR #1026, opened 2020-11-18
WIP to towards better correlation of differential samples

dehann:
towards #1010


dehann:
this work will follow `feat/4Q20/diffjoints`

PR #1027, opened 2020-11-23
Towards removing Graphs.jl for tree

Affie:
- Deprecate TreeClique.index in favor of TreeClique.id
- Fix MetaBayesTree


dehann (approved):

Affie:
This shouldn't be a breaking change, so adding to v0.17.1 as its already merged with other v0.17.1 milestones. Alternatively, we can change all of them to v0.18.0

PR #1028, opened 2020-12-02
Rename softtype -> variableType, getSofttype -> getVariableType

Affie:
This PR 
- makes use of `getVariableType` from DFG instead of `getSofttype` and `.softtype`
- renames and deprecates `TreeBelief` field `softtype` for `variableType`

### DIFF
--- a/src/AnalysisTools.jl
+++ b/src/AnalysisTools.jl
@@ -173,7 +173,7 @@ function mmdSolveKey(vari::DFGVariable,
   tstVal = getBelief(vari, tstKey)
 
   # calc mmd distance
-  mmd(refVal, tstVal, getSofttype(vari), bw=bw)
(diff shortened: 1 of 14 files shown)


dehann:
did the function call deprecation make it into DFG yet, forgot to check?

Affie:
It is deprecated in PR https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/700
`getVariableType` existed before as: `getVariableType(args...) = getSofttype(args...)`
So if function calls are used and not direct field access, we don't need DFG#700

Affie:
Hi @dehann, I think we can do this for v0.17.1 already? That way it is ready for the DFG v0.11 softtype deprecation.

dehann:
alright, sounds good thanks!

Issue #1035, opened 2020-12-08
some graphs cannot be initialized, need warning


dehann:
## Suggestion A

Perhaps terminate the solve with a warning.  Warning rather than hang/crash, allows an automated system to keep trying a solve while the graph is populated until a solve actually occurs.

------

## Issue

CSM is run-away if no initialization is possible in some factor graph fragments that have a tree larger than one clique. eg

```julia
fg = initfg()

addVariable!(fg, :x0, ContinuousScalar)
addVariable!(fg, :x1, ContinuousScalar)
addVariable!(fg, :x2, ContinuousScalar)

addFactor!(fg, [:x0;:x1], LinearRelative(Normal()))
addFactor!(fg, [:x1;:x2], LinearRelative(Normal()))


solveTree!(fg)
```

which results in this:
```julia

julia> solveTree!(fg)
┌ Warning: #TODO Verify useMsgLikelihoods=false works correctly
└ @ IncrementalInference ~/.julia/dev/IncrementalInference/src/SolverAPI.jl:279
[ Info: Ensure variables are all initialized (graphinit)
[ Info: x0 is not initialized, and will do so now...
[ Info: try doautoinit! of x0
[ Info: init with useinitfct Symbol[]
[ Info: x1 is not initialized, and will do so now...
[ Info: try doautoinit! of x1
[ Info: init with useinitfct Symbol[]
[ Info: x2 is not initialized, and will do so now...
[ Info: try doautoinit! of x2
[ Info: init with useinitfct Symbol[]
[ Info: x0 is not initialized, and will do so now...
[ Info: try doautoinit! of x0
[ Info: init with useinitfct Symbol[]
[ Info: x1 is not initialized, and will do so now...
[ Info: try doautoinit! of x1
[ Info: init with useinitfct Symbol[]
[ Info: x2 is not initialized, and will do so now...
[ Info: try doautoinit! of x2
[ Info: init with useinitfct Symbol[]
[ Info: x0 is not initialized, and will do so now...
[ Info: try doautoinit! of x0
[ Info: init with useinitfct Symbol[]
[ Info: x1 is not initialized, and will do so now...
[ Info: try doautoinit! of x1
[ Info: init with useinitfct Symbol[]
[ Info: x2 is not initialized, and will do so now...
[ Info: try doautoinit! of x2
[ Info: init with useinitfct Symbol[]
[ Info: x0 is not initialized, and will do so now...
[...]

PR #1036, opened 2020-12-08
Limit treeinit iters with `limittreeinit_iters` and @error cleanly exit

Affie:
close https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1035

### DIFF
--- a/test/testBasicTreeInit.jl
+++ b/test/testBasicTreeInit.jl
@@ -71,6 +71,35 @@ tree, smt, hist = solveTree!(fg, timeout=70) # , verbose=true, verbosefid=verbos
 
 end
 
+@testset "basic tree initialization limittreeinit_iters" begin
 
+# part of fg that can init
+fg = generateCanonicalFG_lineStep(3; poseEvery=1)
+
+good_vars = sortDFG(ls(fg))
+# fg.solverParams.showtree = true
+# fg.solverParams.drawtree = true
+
+# part of fg that cannot init
+addVariable!(fg, :s0, ContinuousScalar)
+addVariable!(fg, :s1, ContinuousScalar)
+addVariable!(fg, :s2, ContinuousScalar)
+
+addFactor!(fg, [:s0;:s1], LinearRelative(Normal()))
+addFactor!(fg, [:s1;:s2], LinearRelative(Normal()))
+
+smtasks = Task[]
+
+solveTree!(fg; smtasks=smtasks, verbose=true)
+
+
+for var in good_vars
+    sppe = getVariable(fg,var) |> getPPE |> IIF.getPPESuggested
+    println("Testing ", var,": ", sppe)
+    @test isapprox(sppe[1], parse(Int,string(var)[end]), atol=0.1)
+end
+
+
+end
 
 #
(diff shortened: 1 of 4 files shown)


Affie:
I should maybe add test before we merge this. I'll see if I can get to it today.

Affie:
Perhaps we should only increment init counter if `someInit=false`

test/testBasicTreeInit.jl
```diff
@@ -71,6 +71,35 @@ tree, smt, hist = solveTree!(fg, timeout=70) # , verbose=true, verbosefid=verbos
...
+
+smtasks = Task[]
+
+solveTree!(fg; smtasks=smtasks, verbose=true)
+
+
+for var in good_vars
+    sppe = getVariable(fg,var) |> getPPE |> IIF.getPPESuggested
+    println("Testing ", var,": ", sppe)
+    @test isapprox(sppe[1], parse(Int,string(var)[end]), atol=0.1)
```
dehann, line 99: should not be able to extract numerical result because there is no prior anywhere

dehann (changes requested):

dehann:
thanks the counting strategy sounds good

test/testBasicTreeInit.jl
Affie: The test tree looks like this:
image

one part solves fine and is tested numerically

Affie: The test tree looks like this:
image
One part solves fine and has numerical results

dehann:
It's topologically important to have a prior for gauge (absolute reference) rather than vague definition around one variable that happened to have 'initManual!' called.  That kind of information must be done via a Prior

dehann:
besides the isapprox part think PR looks good thanks!

Affie:
On `@errror` vs `@warn`
The current `ERROR` state propagation works with an @error and I didn't want the reason for it hidden as a `@warn`
Here is the output:
See Clique 2 starts the error and its propagated to Clique 4 that only exits because it received an error.
```julia
FSM cliq2, iter=76 -- postUpSolve_StateMachine, NO_INIT
┌ Error: CSM-2e Clique 2 tree init failed, max init retries reached.
└ @ IncrementalInference ~/build/JuliaRobotics/IncrementalInference.jl/src/CliqueStateMachine.jl:424
┌ Error: CSM clique 4: Exit with error state
└ @ IncrementalInference ~/build/JuliaRobotics/IncrementalInference.jl/src/CliqStateMachineUtils.jl:149
FSM cliq4, iter=88 -- exitStateMachine, NO_INIT┌ Error: CSM clique 2: Exit with error state
└ @ IncrementalInference ~/build/JuliaRobotics/IncrementalInference.jl/src/CliqStateMachineUtils.jl:149
FSM cliq2, iter=77 -- exitStateMachine, NO_INIT
[ Info: Finished tree based init-inference
Testing lm0: [0.01460663017623175]
Testing x0: [0.00541170226080372]
Testing x1: [1.0046086867937838]
Testing x2: [1.976521871242177]
Testing x3: [2.9868017889273593]
Test Summary:                                 | Pass  Total
basic tree initialization limittreeinit_iters |    5      5
```

Issue #540, opened 2020-01-16
TreeClique index field and underlining graphs structure indexes


Affie:
### The `TreeClique` `index` field
```julia
mutable struct TreeClique
  index::Int  
  label::Symbol 
  data::Any
  attributes::Dict{String, Any} #The drawing attributes
end
```
#### Incrementally building new trees with clique reuse needs:
- remove clique method (DF agree)
- unique identifier for clique (DF agree, needs to be unique within a USER/ROBOT/SESSION)

Also useful for future distributed trees (cloud) based corresponding concept in DFG.

#### LightGraphs.jl as replacement for Graphs.jl
- For efficiency, vectors of Integers are used for adjacency.  We can extend it and write our own structure but then we have to maintain it separately.
  - DF, I'm apposed to having to maintain more stuff at this point in time.
- the Integers do not maintain any order but stored in a vector.
    - so iteration is easy clIds, ie 1:getNumCliqs(tree)
 
#### Indexes 
There are two places where an `index` / `key` / `id` is needed:
- the `TreeClique` structure
    - To uniquely identify the clique within USER/ROBOT/SESSION
    - _label in DFG_
- the BayesTree structure itself
    - A key value pair, or index to a value, with the value a `TreeClique` 
      - DF, assuming you referring to cliques here, as in `index/key/id => clique container`

#### `index` field
- useful as an unique identifier such as a counter or an autogenerated GUID.
- do not use this `index` field directly as graph index (what is graph index?) as it complicates keeping the this index in sync with the graph (Bayes Tree)
    - can use getIndex(clique::TreeClique) **maybe good way to deprecate**.
      - DF Don't quite understand what you mean here?
- the corresponding concept in DFG is the label that is unique within a graph.
- we should probably rename (`index`) it to avoid confusion. `cliqId`/`id`?

#### Plan is that:
- Integer indexes always refer to the index in the underlying graph.
- `cliq.index` or rather `cliq.id`/`cliq.cliqId` is a unique id within a tree.
  - DF, since I'm confused, yes that would be good idea -- but hold up, way too many things are changing at the same time.  Not quite sure what you working on at the moment -- hence: 

However, cliq.index is very deeply embedded in CSM.
  - Why is this a problem -- I'm a bit confused sorry?
  - a quick search for '.index' gives 235 results, it is also used directly as an index, so CSM will break if you remove a clique.
[...]

Issue #990, opened 2020-10-22
[Feature] Maybe only use clique in CSMC, not full tree


Affie:
Ok, so we can reevaluate in the future:
If we remove the tree from CSMC and only keep a clique, then we can copy the channels from the tree to the CSMC instead of the full parent and child cliques. 
It works by reference so the only difference will be with debugging as deep-copies are made.

_Originally posted by @Affie in https://github.com/JuliaRobotics/IncrementalInference.jl/issues/675#issuecomment-714306518_

dehann:
I think it would be good to ditch csmc.tree if able, as long as we keep the ability to do #742

dehann:
One reason for keeping `csmc.tree` is for debugging, to draw the entire tree state at each step in multithreaded system.  So let's keep tree for now, but what we can do is make sure that we are only using `csmc.clique`, `csmc.parent`, `csmc.children` fields during calculations.  We can avoid pointing back the the `tree` during computations.

This is not a priority now, but later perhaps we can improve the BayesTree api to have only a "shallow" version of the tree for visualization / debugging purposes, much like `DFGVariableSummary`.

PR #1047, opened 2020-12-10
Allow dispatch on cliq id vs graph index by using CliqueId

Affie:
- cleanup deprecated fields `parentCliq` `childCliqs` from `CliqStateMachineContainer`
- change `cliqKey::Int` to  `cliqId::CliqueId{Int}` in `CliqStateMachineContainer`
- remove unused `TreeClique` label field 

This PR is towards having the Tree independent of the underlying graph technology as with DFG.

The idea is that CliqueId is the unique field as per #540

### DIFF
--- a/src/JunctionTree.jl
+++ b/src/JunctionTree.jl
@@ -198,9 +198,6 @@ function appendSeparatorToClique!(bt::AbstractBayesTree, clqID::CliqueId, seprID
   nothing
 end
 
-# FIXME, move Deprecated.jl
-@deprecate appendSeparatorToClique(w...;kw...) appendSeparatorToClique!(w...;kw...)
-
 """
     $SIGNATURES
 
--- a/src/JunctionTree.jl
+++ b/src/JunctionTree.jl
@@ -41,13 +41,15 @@ getFrontals(cliqd::Union{TreeClique,BayesTreeNodeData})::Vector{Symbol} = getCli
 
 Create a new clique.
 """
-function addClique!(bt::AbstractBayesTree,
+function addClique!(bt::MetaBayesTree,
                     dfg::AbstractDFG,
                     varID::Symbol,
                     condIDs::Array{Symbol}=Symbol[]  )
-  bt.btid += 1 #used by Graphs.jl for id -- Graphs.jl will be deprecated
+  bt.btid += 1 #increment cliqueID counter
 
-  clq = TreeClique(bt.btid, string("Clique", bt.btid))
+  cId = CliqueId(bt.btid)
+  
+  clq = TreeClique(cId)
   setLabel!(clq, "")
 
   #TODO addClique!(bt, clq), can't we already have the parent here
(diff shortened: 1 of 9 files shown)


src/JunctionTree.jl
```diff
@@ -184,7 +192,7 @@ end
 
 Add the separator for the newly created clique.
 """
-function appendSeparatorToClique(bt::AbstractBayesTree, clqID::Int, seprIDs::Array{Symbol,1})
+function appendSeparatorToClique(bt::AbstractBayesTree, clqID::CliqueId, seprIDs::Array{Symbol,1})
```
dehann, line 195: Perhaps we should fix the API with a deprecation?
```julia
@deprecate appendSeparatorToClique(w...;kw...) appendSeparatorToClique!(w...;kw...)
```

src/JunctionTreeTypes.jl
```diff
@@ -4,19 +4,7 @@
...
-setUseMetaBayesTree(b::Bool) = global UseMetaBayesTree = b
-function emptyBayesTree()
-  global UseMetaBayesTree
-  if UseMetaBayesTree
-    return MetaBayesTree()
-  else
-    return BayesTree()
-  end
-end
+emptyBayesTree() = MetaBayesTree()
```
dehann, line 7: wonder if we should not include a `const` alias?
```julia
const BayesTree = MetaBayesTree
```

src/JunctionTreeTypes.jl
```diff
@@ -84,8 +72,8 @@ mutable struct CliqStateMachineContainer{BTND, G <: AbstractDFG, InMemG <: InMem
   cliqSubFg::InMemG
   tree::BT
   cliq::TreeClique
-  parentCliq::Vector{TreeClique} #TODO deprecate
-  childCliqs::Vector{TreeClique} #TODO deprecate
+  # parentCliq::Vector{TreeClique} #TODO deprecate
+  # childCliqs::Vector{TreeClique} #TODO deprecate
```
dehann, line 76: I think this is fine for now, but would at some point like to come back and evaluate if it's better in `csmc` to only keep links to `parent` and `children` rather than the entire tree.  That means we would rather do things like:
```julia
getParent(csmc::CliqStateMachineContainer) = csmc.parent
```

This might be important with large distributed trees?

dehann: xref #990

dehann (commented):
I'll see if I can add these few comments to the code quick, thanks for PR!

dehann:
Hi @Affie , would mind reviewing my commits on this PR before you leave for holidays please?

src/JunctionTreeTypes.jl
Affie: Yes, I just wanted to start with a clean slate. Its more to make sure that there is no dead code lingering around.

Affie:
Your changes look good to me. You can merge when the test pass and you are happy.

Affie:
What is the "skip minor cycle" badge? is it for `appendSeparatorToClique` deprecation? Or for my changes? I guess I changed types so it should be minor, even though it's internally used and unexported?

dehann:
> What is the "skip minor cycle" badge

labels are easy to search, so if things break we can see if things like this might be the issue:

> I guess I changed types so it should be minor

minor is a little silly in cases like this one where changes should be completely internal, so the label is just an acknowledgement that we are "skipping" this one.  Who knows if we need to find it again.

Thanks for checking, will merge when ready.  Happy holidays!

src/JunctionTreeTypes.jl
dehann: good plan!

PR #1108, opened 2021-01-04
Switch to Github CI

dehann:
For good example, also see:
https://github.com/mcabbott/Tullio.jl/blob/master/.github/workflows/ci.yml

All this already working on Travis, but new pricing driving us towards Github Actions instead:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/master/.travis.yml

Must do similar for:
- [ ] Caesar ( includes a build and deploy Documentation piece )
- [ ] RoME
- [ ] RoMEPlotting
- [ ] ApproxManifoldProducts
- [ ] DistributedFactorGraphs (also has Docs to build and deploy)
- [ ] KernelDensityEstimate
- [ ] KernelDensityEstimatePlotting

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,29 +43,3 @@ jobs:
       - uses: codecov/codecov-action@v1
         with:
           file: lcov.info
-  docs:
-    name: Documentation
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
-        with:
-          version: '1'
-      - run: |
-          git config --global user.name name
-          git config --global user.email email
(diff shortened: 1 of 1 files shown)


dehann:
Hi @jim-hill-r looks like the existing workflow is already working here.  So think we just have to make sure it runs like CI on master and PR commits and also build / deploy docs on other repos.  Hopefully just a quick thing to finalize.  Thanks for taking a look!

jim-hill-r:
Github has docs for most migrations. Here is the Travis one. https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions. Probably just need to add "on, branch" yaml for branch control. 
image

PR #1139, opened 2021-01-28
Factors return residual for CalcFactorParametric

Affie:


src/Factors/LinearRelative.jl
```diff
@@ -40,14 +40,11 @@ getSample(cf::CalcFactor{<:LinearRelative}, N::Int=1) = (reshape(rand(cf.factor.
...
-                                                  z,
-                                                  x1,
-                                                  x2  ) # where {M<:FactorMetadata,P<:Tuple,X<:AbstractVector}
+function (s::CalcFactor{<:LinearRelative})(z, x1, x2) # where {M<:FactorMetadata,P<:Tuple,X<:AbstractVector}
   #
   # TODO convert to distance(distance(x2,x1),z) # or use dispatch on `-` -- what to do about `.-`
-  res .= z - (x2 - x1)
-  nothing
+  res = z - (x2 - x1)
+  return res
```
dehann, line 47: this looks good to me

Affie: `LinearRelative` is `<:AbstractRelativeRoots` so it needs residual for in-place NLSolve. 
Is the plan to handle the residual in another place then?

dehann: This is where `residual` is currently held (apologies `res` is too short a name, will fix at some point):
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/77233271a32ff14fd8ce2b3cddf0b75decdb88f3/src/FactorGraphTypes.jl#L182

If it makes sense/helps, I have no problem moving `residual` container elsewhere; perhaps into `CalcFactor` or even as high as `CommonConvWrapper` (just remember to respect multithreading -- i.e. `ccw.residual::Vector{RESIDUALTYPE}`, one for each `Threads.nthreads()`.  I'm not sure what the lifespan of `ConvPerThread` will be now that we have introduced `CalcFactor` and `CalcMalahanobisParametric`.  Also have to consolidate `FactorMetadata`, so we expect things to change here over coming weeks. 

My suggestion for this general refactor/consolidation (bigger than just #467) would be small steps via PRs and Review that move quickly rather than trying one big jump.

PR #1140, opened 2021-01-29
Parametric getParametricMeasurement and CalcFactorMahalanobis

Affie:
A step towards consolidation creates residual using `eltype` of variable[1]. 
residual not buffered though, but not to bad.

### DIFF
--- a/src/ParametricUtils.jl
+++ b/src/ParametricUtils.jl
@@ -47,8 +47,8 @@ end
 
 function getParametricMeasurement(Z::Normal)
   meas = mean(Z)
-  σ = 1/std(Z)^2
-  return [meas], reshape([σ],1,1)
+  iσ = 1/std(Z)^2
+  return [meas], reshape([iσ],1,1)
 end
 
 function getParametricMeasurement(Z::MvNormal)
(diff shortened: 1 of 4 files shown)


src/ParametricUtils.jl
```diff
@@ -71,24 +74,64 @@ end
...
+  iΣ::Matrix{Float64}
 end
 
+function CalcFactorMahalanobis(fct::DFGFactor)
+  cf = getFactorType(fct)
+  varOrder = getVariableOrder(fct)
+  meas, iΣ = getParametricMeasurement(cf)
+  calcf = CalcFactor(cf, _getFMdThread(fct), 0, 0, (), [])
+  # TODO figure a way out for type or return residual from factor, see 467
+  res = zeros(Real, length(meas))
```
Affie, line 91: Bad type instability, see discussion in #467

src/ParametricUtils.jl
```diff
@@ -71,24 +74,61 @@ end
...
 # pass in residual for consolidation with nonparametric
 # userdata is now at `cfp.cf.cachedata`
-function (cfp::_CalcFactorParametric)(variables...)
+function (cfp::CalcFactorMahalanobis)(variables...)
   # call the user function (be careful to call the new CalcFactor version only!!!)
-  res = zeros(length(cfp.meanVal))
-  cfp.calcfactor!(res, cfp.meanVal, variables...)
+  res = Vector{eltype(variables[1])}(undef, length(cfp.meas))
+  iΣ = cfp.iΣ
+  cfp.calcfactor!(res, cfp.meas, variables...)
```
dehann, line 98: Yep, this is where the action is at (to consolidate nonparametric and parametric) and looks right to me, thanks!

PS, afterward we can worry about expanding both '(measurement..., variables...)'

dehann (approved):

PR #1147, opened 2021-02-05
Return Residual in CalcFactor

Affie:
WIP towards #467

### DIFF
--- a/src/DeconvUtils.jl
+++ b/src/DeconvUtils.jl
@@ -90,7 +90,7 @@ function approxDeconv(fcto::DFGFactor,
     #
     
     # lambda with which to find best measurement values
-    hypoObj = (res, tgt) -> (targeti_.=tgt; onehypo!(res) )
+    hypoObj = (tgt) -> (targeti_.=tgt; onehypo!() )
 
     # find solution via SubArray view pointing to original memory location
     targeti_ .= _solveLambdaNumeric(fcttype, hypoObj, res_, measurement[1][:,idx], islen1)
--- a/src/NumericalCalculations.jl
+++ b/src/NumericalCalculations.jl
@@ -43,7 +43,7 @@ function _solveLambdaNumeric( fcttype::Union{F,<:Mixture{N_,F,S,T}},
   #
 
   #
-  r = nlsolve(objResX, u0, inplace=true)
+  r = nlsolve( (res, x) -> res .= objResX(x), u0, inplace=true)
 
   #
   return r.zero
@@ -58,9 +58,9 @@ function _solveLambdaNumeric( fcttype::Union{F,<:Mixture{N_,F,S,T}},
   #
   # wrt #467 allow residual to be standardize for Roots and Minimize and Parametric cases.
   r = if islen1
-    optimize((x) -> (objResX(residual, x); sum(residual.^2)), u0, BFGS() )
+    optimize((x) -> (residual = objResX(x); sum(residual.^2)), u0, BFGS() )
   else
-    optimize((x) -> (objResX(residual, x); sum(residual.^2)), u0)
+    optimize((x) -> (residual = objResX(x); sum(residual.^2)), u0)
   end
 
   # 
@@ -148,7 +148,7 @@ function _buildCalcFactorLambdaSample(ccwl::CommonConvWrapper,
   fill!(cpt_.res, 0.0) # Roots->xDim | Minimize->zDim
 
   # build static lambda
-  unrollHypo! = (res) -> cf( res, (_viewdim1or2.(measurement_, :, smpid))..., (view.(varParams, :, smpid))... )
+  unrollHypo! =  ()->cf( (_viewdim1or2.(measurement_, :, smpid))..., (view.(varParams, :, smpid))... )
 
   return unrollHypo!, target
 end
@@ -195,7 +195,7 @@ function _solveCCWNumeric!( ccwl::Union{CommonConvWrapper{F},
   
   # broadcast updates original view memory location
   ## using CalcFactor legacy path inside (::CalcFactor)
-  _hypoObj = (res, x) -> (target.=x; unrollHypo!(res) )
+  _hypoObj = (x) -> (target.=x; unrollHypo!())
   
   # TODO small off-manifold perturbation is a numerical workaround only, make on-manifold requires RoME.jl #244
   # use all element dimensions : ==> 1:ccwl.xDim
--- a/src/ParametricUtils.jl
+++ b/src/ParametricUtils.jl
@@ -40,8 +40,8 @@ end
     $SIGNATURES
 
 Returns the parametric measurement for a factor as a tuple (measurement, inverse covariance) for parametric inference (assumign Gaussian).
-Defaults to find the parametric measurement at field `Z` followed by `z`.  
-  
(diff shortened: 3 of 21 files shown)


src/NumericalCalculations.jl
```diff
@@ -195,7 +195,7 @@ function _solveCCWNumeric!( ccwl::Union{CommonConvWrapper{F},
   
   # broadcast updates original view memory location
   ## using CalcFactor legacy path inside (::CalcFactor)
-  _hypoObj = (res, x) -> (target.=x; unrollHypo!(res) )
+  _hypoObj = (x) -> (target.=x; unrollHypo!())
```
dehann, line 198: this has to remain `(res, x) ->` as standard for both Optim and NLsolve, and is used as such by both `approxConv` and `approxDeconv`

src/ParametricUtils.jl
```diff
@@ -102,13 +102,10 @@ end
 
 # This is where the actual parametric calculation happens, CalcFactor equivalent for parametric
 function (cfp::CalcFactorMahalanobis)(variables...)
-  #TODO res might be removed see https://github.com/JuliaRobotics/IncrementalInference.jl/issues/467#issuecomment-772987921
-  res = Vector{eltype(variables[1])}(undef, length(cfp.meas))
-  iΣ = cfp.iΣ
-  cfp.calcfactor!(res, cfp.meas, variables...)
-  
+  # call the user function (be careful to call the new CalcFactor version only!!!)
+  res = cfp.calcfactor!(cfp.meas, variables...)
```
dehann, line 106: Also expand measurements if more than one (`getSample` always returns a `::Tuple`):
```julia
res = cfp.calcfactor!(cfp.meas..., variables...)
```

dehann (changes requested):

src/NumericalCalculations.jl
dehann: https://github.com/JuliaRobotics/IncrementalInference.jl/blob/648aa8ca63eaadd69631b1532bc30cfd96efdb68/src/DeconvUtils.jl#L93

dehann: https://github.com/JuliaRobotics/IncrementalInference.jl/blob/648aa8ca63eaadd69631b1532bc30cfd96efdb68/src/NumericalCalculations.jl#L46

dehann: https://github.com/JuliaRobotics/IncrementalInference.jl/blob/648aa8ca63eaadd69631b1532bc30cfd96efdb68/src/NumericalCalculations.jl#L63

src/ParametricUtils.jl
Affie: This is the measurement from `getParametricMeasurement` it currently returns a `Vector{Float64}`. In what cases would a parametric solve be able to handle more than this? An example or use case would be helpful for development and testing

src/NumericalCalculations.jl
Affie: I don't completely follow, It didn't look like the residual was used anywhere, so I just took it out. 
The residual here: 
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/583a1ca2c468775e2e36cb7eb079f17bb8e8ddaa/src/NumericalCalculations.jl#L40
and here:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/583a1ca2c468775e2e36cb7eb079f17bb8e8ddaa/src/NumericalCalculations.jl#L54
is not populated or used

dehann: I think you might not be looking at the code carefully enough.  To ensure the higher up calls are identical, each Roots and Minimize case has one redundant definition of the `residual`.  Thats why in some cases the parameter `residual` might look excessive.

-------

Look more closely at the Minimize snippet you posted.  The `residual` parameter **is being used** on line 61
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/583a1ca2c468775e2e36cb7eb079f17bb8e8ddaa/src/NumericalCalculations.jl#L54-L61

as I tried to show here: https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1147#discussion_r571094577

Is it necessary to drop the `::AbstractVector{<:Real}` type restriction for autodiff to work?

-----

In the Roots case, the lambda parameter `res` is a **requirement** from NLsolve.jl package, as I tried to show here:
https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1147#discussion_r571093888

src/ParametricUtils.jl
dehann: okay, lets leave as you have it for now.

src/NumericalCalculations.jl
dehann: perhaps we should add the comment:
```julia
# for code symmetry between Minimize (Optim.jl) and Roots (NLsolve.jl), we use standardized lambda `(res, x)-> obj...`
```

src/DeconvUtils.jl
```diff
@@ -90,7 +90,7 @@ function approxDeconv(fcto::DFGFactor,
     #
     
     # lambda with which to find best measurement values
-    hypoObj = (res, tgt) -> (targeti_.=tgt; onehypo!(res) )
+    hypoObj = (tgt) -> (targeti_.=tgt; onehypo!() )
```
dehann, line 93: i missed this before, same deal: we need standardized lambda interface `(res,x)->obj...`

src/NumericalCalculations.jl
```diff
@@ -43,7 +43,7 @@ function _solveLambdaNumeric( fcttype::Union{F,<:Mixture{N_,F,S,T}},
   #
 
   #
-  r = nlsolve(objResX, u0, inplace=true)
+  r = nlsolve( (res, x) -> res .= objResX(x), u0, inplace=true)
```
dehann, line 46: Ah, you are doing the opposite of what I was doing.  I had `residual` explicitly available for in-place operations (which we will still keep somewhere in `CalcFactor.residual::T` for hard in-place factor situations).

I see here now that you are changing to a more implicit usage of `residual` in the Roots (NLsolve.jl) case.  I did not see this before, sorry if I missed it.

src/NumericalCalculations.jl
dehann: i think the part I was missing earlier is you added residual in a new lambda at `nlsolve` instead.  I commented separately there a minute ago

dehann (approved):
https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1147/files#r571239052

src/DeconvUtils.jl
dehann: ignore

src/NumericalCalculations.jl
```diff
@@ -58,9 +58,9 @@ function _solveLambdaNumeric( fcttype::Union{F,<:Mixture{N_,F,S,T}},
   #
   # wrt #467 allow residual to be standardize for Roots and Minimize and Parametric cases.
   r = if islen1
-    optimize((x) -> (objResX(residual, x); sum(residual.^2)), u0, BFGS() )
+    optimize((x) -> (residual = objResX(x); sum(residual.^2)), u0, BFGS() )
```
Affie, line 61: Reading your comments I saw it looks like i forgot the `.=`

Affie: #1150

PR #1150, opened 2021-02-06
Update residual in _solveLambdaNumeric optim

Affie:
I missed the dots. I kept missing it with CalcFactos also.

### DIFF
--- a/src/NumericalCalculations.jl
+++ b/src/NumericalCalculations.jl
@@ -58,9 +58,9 @@ function _solveLambdaNumeric( fcttype::Union{F,<:Mixture{N_,F,S,T}},
   #
   # wrt #467 allow residual to be standardize for Roots and Minimize and Parametric cases.
   r = if islen1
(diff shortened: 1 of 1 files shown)


dehann (approved):
hi, should be fine to broadcast thanks!

Affie:
It's hard to know what will be faster without benchmarking. Since optim works with a single scalar cost variable, I dought it will make any difference. 
That being said, I think there is still a place for `CalcFactor.residual`, but for more specialised cases.

Affie:
julia 1.6 beta test only fails because of random numerical test failure. 1.5 passes, so will merge.

dehann:
HI @Affie , just to reaffirm this, add a +1 or -1 when you read this

> I think there is still a place for CalcFactor.residual, but for more specialised cases.

Agree, I think there will be specialized cases where in-place memory for `residual` will be needed.  Understand the type definition on `residual` is only near runtime.  I think this will be easier to sort out once CPT, CCW, FMd, CalcFactor are better consolidated and all factors are converted to the single definition.

PR #1154, opened 2021-02-09
Parametric solve updates for CalcFactor 

Affie:
- [x] Remove old parametric factor functions


src/ParametricUtils.jl
```diff
@@ -138,20 +138,20 @@ function _totalCost(cfdict::Dict{Symbol, CalcFactorMahalanobis},
 end
 
 
-# build the cost function
+# DEPRECATED slower version
```
dehann, line 141: Could we move this function to the Deprecated.jl file perhaps?

Affie: I didn't because it's still in use and I'm not 100% sure in which direction to go. The overhead bothers me a bit and I think it could be better to add some parametric stuff to something like CCW.

Affie:
Will merge this as is and continue with CCW in a separate PR.

Issue #1051, opened 2020-12-14
Proposal distribution inflation


dehann:
### Summary

added
```julia
fg = initfg()
getSolverParams(fg).inflation = 3.0 # the default

# and per factor control
addFactor!(fg, ...., inflation=0.01) # or 0 to switch off like before
```
------

### Original Comment

It might be worth experimenting with an option to inject more entropy when calculating proposal distributions before the product step.  Likely best to start with default off, but mathematically not expected bias the posterior estimates but rather enhance the reach over the domains when searching for a posterior.

dehann:
xref JuliaRobotics/RoME.jl#380

dehann:
Suggestion is to generalize this:

https://github.com/JuliaRobotics/IncrementalInference.jl/blob/c6a845763448af40983755449e245d2a7e8b41dc/src/ApproxConv.jl#L399

will fix less important "hack" aspect after JuliaRobotics/RoME.jl#244 and JuliaRobotics/ApproxManifoldProducts.jl#41 are complete.

dehann:
decision on unconstrained dimensions with only a partial constraint, whether noise should be added or not.  Might be useful as an option that can be turned on or off, and dependent on mechanics inside inference.  Perhaps start with default off with intention to switch on when ready.

dehann:
Hi @Affie , just an FYI that to fix this I have to add `DFG.GenericFunctionNodeData.inflation`.  Will link separate PR here, and likely need to tag DFG v0.12.0 as a result.

dehann:
automation closed this too soon

dehann:
Likely have to look at relative entropy injection per dimension.  e.g. Pose3 test case in RoME results in double peaks on rotations:
![Screenshot from 2021-02-13 23-15-05](https://user-images.githubusercontent.com/6412556/107868313-7c322700-6e51-11eb-9851-c255a5dd04af.png)

investigating more.

PR #1158, opened 2021-02-12
multihypo fix and simple test

Affie:
@dehann, just check if I didn't make any mistakes, it's late.


dehann:
Hi Johan, will do thanks -- wow there has been so much refactoring and internal consolidation.  Hope we did not miss something in the tests.  Thanks for adding a new test to help strengthen that!

dehann:
Will work on this for IIF v0.21.2, but do want to get the fix for #1051 out sooner.  Plan is to tag v0.21.1 and then return to this issue.

dehann:
Hi @Affie ,

Just an FYI on improved results on previous `testMultimodal1D.jl` where we had `:lm1, :lp1` and `:lm2, :lp2`.  We can now get more better multimodal results over all several variables:
Screenshot from 2021-02-22 20-58-08

dehann:
together with previous fixes for RoME 380, this should fix JuliaRobotics/RoME.jl#378

PR #1159, opened 2021-02-14
experimental inflation, #1051

dehann:
xref #1051


dehann:
xref JuilaRobotics/RoME.jl#380

PR #1167, opened 2021-02-16
double direct during Gibbs

dehann:
xref 
- JuliaRobotics/RoME.jl#380 and 
- JuliaRobotics/RoME.jl#416


dehann:
after #1168 see here for which tests should be updated

Issue #1168, opened 2021-02-16
`solveCliq!` tripped up by tree `:upMsg` error


dehann:
```julia
N=100
fg = initfg()

addVariable!(fg, :x0, ContinuousEuclid{2}, N=N)
addFactor!(fg, [:x0], Prior(MvNormal([100.0;0], [1;1.0])), graphinit=false)

addVariable!(fg, :x1, ContinuousEuclid{2}, N=N)
addFactor!(fg, [:x1], Prior(MvNormal([0.0;100.0], [1;1.0])), graphinit=false)

addVariable!(fg, :l1, ContinuousEuclid{2}, N=N)
addFactor!(fg, [:x0;:l1], EuclidDistance(Normal(100.0, 1.0)), graphinit=false)
addFactor!(fg, [:x1;:l1], EuclidDistance(Normal(100.0, 1.0)), graphinit=false)

## 

eo = [:x1; :x0; :l1]
tree = buildTreeReset!(fg, eo)

stuff = solveCliq!(fg, tree, :x1)
```

Produces error:
```julia
ERROR: KeyError: key :upMsg not found
Stacktrace:
  [1] getindex
    @ ./dict.jl:482 [inlined]
  [2] get_prop
    @ ~/.julia/packages/MetaGraphs/SUjFO/src/MetaGraphs.jl:256 [inlined]
  [3] getMsgUpChannel
    @ ~/.julia/dev/IncrementalInference/src/TreeMessageAccessors.jl:63 [inlined]
  [4] putBeliefMessageUp!(tree::MetaBayesTree, edge::LightGraphs.SimpleGraphs.SimpleEdge{Int64}, beliefMsg::LikelihoodMessage{IncrementalInference.NonparametricMessage})
    @ IncrementalInference ~/.julia/dev/IncrementalInference/src/TreeMessageAccessors.jl:72
  [5] postUpSolve_StateMachine(csmc::CliqStateMachineContainer{BayesTreeNodeData, LightDFG{SolverParams, DFGVariable, DFGFactor}, LightDFG{SolverParams, DFGVariable, DFGFactor}, MetaBayesTree})
    @ IncrementalInference ~/.julia/dev/IncrementalInference/src/CliqueStateMachine.jl:437
[...]

PR #1169, opened 2021-02-17
solveCliq! in isolation

Affie:
Hi @dehann, I don't know if this is what you had in mind for `solveCliqu!` from #1168. 

I didn't know exactly what we wanted to do with solveCliq!. Problem was: It is still solving as if its part of the larger tree, so it's trying to access messages from the rest of the tree. It's easy to add the channels for the messages but then it would just block on them waiting. So where should these messages come from?

For this PR I went with solving the clique in isolation, it doesn't make sense if its not a leave clique though.


dehann (approved):
Great thanks!

PR #1173, opened 2021-02-19
Euclid Distance fg generator and tests

Affie:
@dehann, I just want to commit what I have so long on tests. I got a bit distracted with debugging and stripped out the test separately.

### DIFF
--- a/src/CanonicalGraphExamples.jl
+++ b/src/CanonicalGraphExamples.jl
@@ -169,3 +169,36 @@ function generateCanonicalFG_lineStep(  lineLength::Int;
 
     return fg
 end
+
+export generateCanonicalFG_EuclidDistance
+"""
+    $SIGNATURES
+Generate a EuclidDistance test graph where 1 landmark position is unknown. 
+"""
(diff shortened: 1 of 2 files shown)


dehann:
HI Johan, thanks this is a much cleaner way to build the tests.  I'm going to merge so long and continue work on fixing the "maybe Optim zero preference" issue we discussed.  Please continue to develop any further tests as you intended on this branch.  I will work separately.

Affie:
Hi, will do. I’m sure you saw the 1d example `@test_broken` it behaves the same as 2 d 2 points and one landmark, but it’s a bit easier to debug.

dehann:
I did see the 1D broken examples thanks, and the best thing to do there at this time is push the inflation really high, because the proposal distribution modes in this case are disconnected (not a conjoined ring).  So it is also an example of where the solution need to get bumped out of a local region.  I did find that making the inflation number high in general results in lots of poor parasitic performance.  Therefore left these new tests broken for now, and will pick up again at some point after #1010 is done.

PR #1181, opened 2021-02-23
Update ci.yml

Affie:
build some branches


Affie:
Failed slight numerical test on only 1.6, so merging

PR #1186, opened 2021-02-25
add solveCliqDown!

### LINKED ISSUE #1170: solveCliq feature update
Just documenting what we spoke about...
- [x] record cliques replace old solveCliq function
- [x] solve clique will do up/down
- [x] in messages can be created from fg or input directly
- [x] out messages will be output
- [x] EDIT, DF, option to bring out CSM history vector for step by step analysis.
- therefore, it becomes more of a debug feature on how a clique is solved with different messages
- we can add extra plumbing later to ease clique interaction.

Affie:
close #1170

### DIFF
--- a/src/SolverAPI.jl
+++ b/src/SolverAPI.jl
@@ -253,7 +253,7 @@ tree, smt, hist = solveTree!(fg [,tree])
 
 Related
 
-[`solveCliq!`](@ref), [`buildTreeReset!`](@ref), [`repeatCSMStep`](@ref), [`printCSMHistoryLogical`](@ref)
+[`solveCliqUp!`](@ref), [`solveCliqDown!`](@ref), [`buildTreeReset!`](@ref), [`repeatCSMStep`](@ref), [`printCSMHistoryLogical`](@ref)
 """
 function solveTree!(dfgl::AbstractDFG,
                     oldtree::AbstractBayesTree=BayesTree();
--- a/src/SolverAPI.jl
+++ b/src/SolverAPI.jl
@@ -410,76 +410,15 @@ See `solveTree!`.
 """
 const solveGraph! = solveTree!
 
-"""
-    $SIGNATURES
-
-Perform inference over one clique in the Bayes tree according to `opt::SolverParams`.
-
-Example
-```julia
-tree = buildTreeReset!(fg)
-hist = solveCliq!(fg, tree, :x1, recordcliq = true )
-
-# print CSM steps
-printCliqHistorySummary(hist)
-```
-
-DevNotes
-- on `sandboxStateMachineStep`, see FSM #42
-
-Related
-
-[`solveTree!`](@ref), [`buildTreeReset!`](@ref), [`printCliqHistorySummary`](@ref), [`repeatCSMStep!`](@ref), `sandboxStateMachineStep`
-"""
-function solveCliq!(dfgl::AbstractDFG,
-                    tree::AbstractBayesTree,
-                    cliqid::Symbol;
-                    verbose::Bool=false,
-                    recordcliq::Bool=false,
-                    # cliqHistories = Dict{Int,Vector{CSMHistoryTuple}}(),
-                    async::Bool=false )
-  #
-  # hist = Vector{CSMHistoryTuple}()
-  opt = DFG.getSolverParams(dfgl)
-
-  if opt.isfixedlag
-      @info "Quasi fixed-lag is enabled (a feature currently in testing)!"
-      fifoFreeze!(dfgl)
-  end
-
-  # if !isTreeSolved(treel, skipinitialized=true)
-  cliq = getClique(tree, cliqid)
-
-  # modyfy local copy of the tree
-  tree_ = deepcopy(tree)
-  # isolate clique
-  deleteClique!(tree_, getParent(tree_, cliq)[1])
-  foreach(c->deleteClique!(tree_,c), getChildren(tree_, cliq))
-
-  cliqtask = if async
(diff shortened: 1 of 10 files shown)


src/SolverAPI.jl
```diff
@@ -558,3 +563,79 @@ function solveCliqUp!(fg::AbstractDFG,
 
   return hist, beliefMessageOut
 end
+
+"""
+Internal function used for solveCliqDown!
```
dehann, line 568: Perhaps just specify Tx or Rx at this `cliq` vs `parent`

src/SolverAPI.jl
```diff
@@ -558,3 +563,79 @@ function solveCliqUp!(fg::AbstractDFG,
...
+"""
+function _buildMessageDown(fg, tree, cliqid)
+  cliq = getClique(tree, cliqid)
+  parent = getParent(tree, cliq)[1]
+  return getCliqDownMsgsAfterDownSolve(fg, parent; status=DOWNSOLVED)
```
dehann, line 573: what if the clique has status `MARGINALIZED`, then `DOWNSOLVED` is not entirely correct.  Is there perhaps an easy way to use the actual clique status vs hardcoded?

dehann: or via keyword to allow user to override the default status?

src/SolverAPI.jl
```diff
@@ -558,3 +563,79 @@ function solveCliqUp!(fg::AbstractDFG,
...
+      messages[e.dst] = takeBeliefMessageDown!(tree, e) 
+    end
+    messages
+  end
+
+  recordcliqs = recordcliq ?  [getFrontals(cliq)[1]] : Symbol[]
+
+  hist = tryCliqStateMachineSolve!(fg, tree, cliq.id; 
+                                   verbose=verbose, drawtree=opt.drawtree, limititers=opt.limititers,
+                                   recordcliqs=recordcliqs, incremental=opt.incremental)
```
dehann, line 632: in future versions we can bring back the `async` option.  Just keeping a note here.

src/TreeMessageUtils.jl
```diff
@@ -574,7 +574,7 @@ during clique inference.
 
 DevNotes
 - TODO make `::Vector{Symbol}` version.
-
+- TODO function taking fcts::Vector{DFGFactor} is unused and replace by the tags version, perhaps we can remove it.
```
dehann, line 577: just making sure, `tags` or `labels`?

src/SolverAPI.jl
dehann: also e.g.:
https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1186/files#diff-bdbb86421fa7a813967257eb8845a821e8595683ead3f4615750646db435d942R640

src/TreeMessageUtils.jl
Affie: tags, If you remember we remove messages based on the `__LIKELIHOOD_MESSAGE__` tag

dehann (approved):
Looks good thanks.  Perhaps just take a look at the comments and see if any are useful or valid.

dehann:
CI on 1.5 failed due to numerical error `(1.15 < 1.0 )`.  I restarted it.  The new multihypo updates (although not included here yet) might fail on old obsolete tests, and there are too many to search so I'm checking all CI outcomes instead.

src/SolverAPI.jl
Affie: There is no parent clique, so no easy way. 
We can do kwarg, but does it matter? Downsolve is treated the same in all cases.

dehann: i'd suggest just going with kwarg to make the hard coded a little better.  We'll see if tree CSM logic evolves to a point where DOWNSOLVE vs MARGINALIZED or others make a difference -- agreed not 100% yet today, but I have reason to suspect things will change there over time.

src/SolverAPI.jl
Affie: I'm not following this comment, do you want clearer comments on whether it is an out or in message?

Affie: put is Tx and take! is Rx

src/SolverAPI.jl
```diff
@@ -410,73 +410,16 @@ See `solveTree!`.
...
-printCliqHistorySummary(hist)
-```
-
-DevNotes
-- on `sandboxStateMachineStep`, see FSM #42
-
-Related
-
-[`solveTree!`](@ref), [`buildTreeReset!`](@ref), [`printCliqHistorySummary`](@ref), [`repeatCSMStep!`](@ref), `sandboxStateMachineStep`
+Internal function used for solveCliqUp! to build the incomming upward message (Rx)
```
dehann, line 416:
```suggestion
Internal function used for solveCliqUp! to build the incoming upward message (Rx)
```

src/SolverAPI.jl
```diff
@@ -558,3 +504,78 @@ function solveCliqUp!(fg::AbstractDFG,
 
   return hist, beliefMessageOut
 end
+
+"""
+    $SIGNATURES
+Internal function used for solveCliqDown! to build the incomming downward message (Rx)
```
dehann, line 510:
```suggestion
Internal function used for solveCliqDown! to build the incoming downward message (Rx)
```

Affie: Thanks, I need a spell checker...

Affie:
Test passed, just spelling from code review. So not waiting for the new test.

PR #1226, opened 2021-04-12
towards solveTree! with unique solveKey

dehann:
MWE for #1219 
```julia
solveTree!(fg, solveKey=:greatscott21)
```

cc @jim-hill-r , work in progress to make this possible.


dehann:
xref #1229

GearsAD:
🎉

dehann:
HI @jim-hill-r , im merging this.  Might still have a bug or two.  I'm at the point of diminishing returns and would rather keep going on getting you test scripts and find bugs in aggregate.  

I did have to break something small with parametric solve, but the particular API is still in development and asked @Affie to separately include that with his ongoing work on finishing the parametric consolidation.  That issue can be followed at #1229  if interested.

dehann:
THis will be available in v0.23.2

PR #1276, opened 2021-06-18
Sandbox optimization on manifold

Affie:
Here is my sanbox on manifold optimization. I updated an old attempt on Optim.jl to use un/coords and actually got it running.


dehann (approved):
Looks good to me thanks!  flatten might the better way to transition the serialization stuff for Manifolds.jl types

Affie:
I tested at pi and it's not working properly there.

Affie:
I'm merging this, just to have it in the latest master to work from. It can be deleted later.

dehann:
sounds good thanks, this sandbox is going to help a lot!

PR #1283, opened 2021-06-28
update for val::Vector{P<:AbstractVector{<:Real}}

dehann:
Needs 
- DFG v0.15.0
- AMP v0.4.0


dehann:
xref JuliaManifolds/Manopt.jl#90

dehann:
HI @Affie , I'm going to merge this so both of us can work against master.  I'm pretty something is broken since the numerical tolerances seem slightly worse after this PR, however, when I test each individually they seem to run through.  So, my plan is not to tag IIF until we have made it through the tests on RoME with this new Manifolds.jl approach, including this ::Vector{P} upgrade.

PR #1291, opened 2021-07-05
wip, consolidate factors and manifolds

dehann:
I'd like to consolidate ManifoldPrior with Prior, but this will need:
- converters to serialize manifolds (I was thinking of just doing it by name , ie shorthand string, and hard coding the converters, one for each manifold).  e.g. "SpecialEuclidean(3)" gets persisted, and user code converts it back to whatever complicated type.
- helper constructors so that legacy use of say `Prior(Normal())` automatically inserts `M=Euclidean(1)`, 

EDIT @Affie , feel free to take over this branch in the morning hours there.  I will be working on this again tomorrow.


Affie:
The manifolds prior part of this is currently in conflict with local changes I have as a wip to allow parametric and non-parametric to work with the same factors. 
Can we hold off on consolidating Prior and ManifoldPrior for now?
I think we can serialize in the same way as with variables with a `getManifold` function if it's easier, but it is currently more convenient to just store the manifold in the factor.

dehann:
okay, sure holding off on consolidation of Prior.  I'm going to clear vecP stuff in RoME in the meantime then.

Affie:
Hi, I've uploaded my local changes as is so long: #1292

dehann:
Replaced by 0f7edd4 in #1300

Issue #1280, opened 2021-06-20
on-manifold covariance estimation


dehann:
upgrade this:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/7f45f6317060e6cdb577358624369c1a8bb677f1/src/ApproxConv.jl#L156

Affie:
I didn't realize there was an issue on this one, here is what i have so far: https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1292/commits/5adf2f3b369e45d25dd3bd18f19d7d1ebe89eb55

source https://hal.inria.fr/inria-00614994/PDF/Pennec.JMIV06.pdf

PR #1292, opened 2021-07-05
WIP on manifold optimization and factors

Affie:
- Factors returns tangent vector at p
- _solveLambdaNumeric does distance2
- parametric does Mahalanobis distance2

Tests added:
- @test_broken that fails on Sphere(2) - #1293
- test on SO(2)
- test on SO(3)

Needs:
- https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/94
- ~~https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/788~~
    - we are storing coordinates for now

Breaks (TODO):
- ~~Maybe CCW serialization~~ - slightly related to https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/590#issuecomment-776838053
- Skips compare on `vartypes` field in ccw. captured in https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/434
- ~~Parametric~~

### DIFF
--- a/examples/dev/testManiFactors.jl
+++ b/examples/dev/testManiFactors.jl
@@ -3,6 +3,24 @@ using IncrementalInference
 using StaticArrays
 using Manifolds
 
+
+##
+
+
+
+fg = initfg()
+
+v0 = addVariable!(fg, :x0, Pose2)
+v1 = addVariable!(fg, :x1, Point2)
+v2 = addVariable!(fg, :x2, Pose2)
+
+fg.solverParams.graphinit = false
+
+addFactor!(fg, [:x0], Prior(Normal()))
+addFactor!(fg, [:x0,:x1], Prior(Normal()))
+addFactor!(fg, [:x1,:x2], Prior(Normal()))
+addFactor!(fg, [:x0,:x2], Prior(Normal()))
+
 ## ======================================================================================
 ## 
 ## ======================================================================================
@@ -17,7 +35,7 @@ v0 = addVariable!(fg, :x0, Pose2)
 v1 = addVariable!(fg, :x1, Pose2)
 
 
-mp = ManifoldPrior(SpecialEuclidean(2), MvNormal([1.0, 1.0, 0.0]), ProductRepr(SA[0., 0], SA[1.0 0; 0 1]))
+mp = ManifoldPrior(SpecialEuclidean(2),ProductRepr(SA[0., 0], SA[1.0 0; 0 1]), MvNormal([1.0, 1.0, 0.0]))
 p = addFactor!(fg, [:x0], mp;  graphinit=true)
 
 mf = ManifoldFactor(SpecialEuclidean(2), MvNormal([0.1, 0.2, 0.01]))
@@ -44,7 +62,7 @@ v0 = addVariable!(fg, :x0, Point2)
 v1 = addVariable!(fg, :x1, Point2)
 
 
-mp = ManifoldPrior(TranslationGroup(2), MvNormal([10.0, 20.0], [1.0,1.0]), SA[0., 0])
+mp = ManifoldPrior(TranslationGroup(2), SA[10., 20], MvNormal([1.0,1.0]))
 p = addFactor!(fg, [:x0], mp;  graphinit=true)
 
 doautoinit!(fg, :x0)
@@ -62,7 +80,7 @@ solveGraph!(fg)
 ## 
 ## ======================================================================================
 
-# Base.convert(::Type{<:Tuple}, M::SpecialOrthogonal{2}) = (:Circular,)
+Base.convert(::Type{<:Tuple}, M::SpecialOrthogonal{2}) = (:Circular,)
 
 @defVariable SO2 SpecialOrthogonal(2) [0.0]
 getManifold(SO2)
@@ -75,7 +93,7 @@ v0 = addVariable!(fg, :x0, SO2)
 v1 = addVariable!(fg, :x1, SO2)
 
 
-mp = ManifoldPrior(SpecialOrthogonal(2), MvNormal([1.0]), SA[1.0 0.0; 0 1])
+mp = ManifoldPrior(SpecialOrthogonal(2), SA[1.0 0.0; 0 1], MvNormal([1.0]))
 p = addFactor!(fg, [:x0], mp;  graphinit=true)
 
 doautoinit!(fg, :x0)
@@ -109,11 +127,11 @@ v0 = addVariable!(fg, :x0, Sphere2)
 v1 = addVariable!(fg, :x1, Sphere2)
 
 
-mp = ManifoldPrior(Sphere(2), MvNormal([0.0, 0.0], [0.01, 0.01]), SA[1., 0, 0])
+mp = ManifoldPrior(Sphere(2), SA[1., 0, 0], MvNormal([0.01, 0.01]))
 p = addFactor!(fg, [:x0], mp)
 
 doautoinit!(fg, :x0)
-@enter doautoinit!(fg, :x0)
+# @enter doautoinit!(fg, :x0)
 
 # @enter addFactor!(fg, [:x0], mp;  graphinit=true)
 
--- a/examples/dev/testManiFactors.jl
+++ b/examples/dev/testManiFactors.jl
@@ -3,47 +3,6 @@ using IncrementalInference
 using StaticArrays
 using Manifolds
 
-
-##
-
-
-
-fg = initfg()
-
-v0 = addVariable!(fg, :x0, Pose2)
-v1 = addVariable!(fg, :x1, Point2)
-v2 = addVariable!(fg, :x2, Pose2)
-
-fg.solverParams.graphinit = false
-
-addFactor!(fg, [:x0], Prior(Normal()))
-addFactor!(fg, [:x0,:x1], Prior(Normal()))
-addFactor!(fg, [:x1,:x2], Prior(Normal()))
-addFactor!(fg, [:x0,:x2], Prior(Normal()))
-
-## ======================================================================================
-## 
-## ======================================================================================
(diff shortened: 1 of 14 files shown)


examples/dev/testManiFactors.jl
```diff
@@ -3,6 +3,24 @@ using IncrementalInference
...
+v0 = addVariable!(fg, :x0, Pose2)
+v1 = addVariable!(fg, :x1, Point2)
+v2 = addVariable!(fg, :x2, Pose2)
+
+fg.solverParams.graphinit = false
+
+addFactor!(fg, [:x0], Prior(Normal()))
+addFactor!(fg, [:x0,:x1], Prior(Normal()))
+addFactor!(fg, [:x1,:x2], Prior(Normal()))
+addFactor!(fg, [:x0,:x2], Prior(Normal()))
```
dehann, line 22: Regular use is `PriorPose2` `PriorPoint2` etc, don't worry about trying to make all work through just generic `Prior`

examples/dev/testManiFactors.jl
```diff
@@ -93,8 +111,8 @@ f1 = addFactor!(fg, [:x0,:x1])
 ## Sphere(n)
 ## ======================================================================================
 
-Base.convert(::Type{<:Tuple}, M::Sphere{2, ℝ}) = (:Euclid, :Euclid, :Euclid)
-Base.convert(::Type{<:Tuple}, ::IIF.InstanceType{Sphere{2, ℝ}})  = (:Euclid, :Euclid, :Euclid)
+Base.convert(::Type{<:Tuple}, M::Sphere{2, ℝ}) = (:Euclid, :Euclid)
+Base.convert(::Type{<:Tuple}, ::IIF.InstanceType{Sphere{2, ℝ}})  = (:Euclid, :Euclid)
 
 @defVariable Sphere2 Sphere(2) [1.0, 0.0, 0.0]
```
dehann, line 117: don't worry about Sphere(2) now if it is not compatible with current KDE.  That will later be replaced by AMP.  See JuliaRobotics/ApproxManifoldProducts.jl#41

src/Factors/GenericFunctions.jl
```diff
@@ -119,7 +124,8 @@ function getSample(cf::CalcFactor{<:ManifoldPrior}, N::Int=1)
     
     Xc = [rand(Z) for _ in 1:N]
     
-    X = get_vector.(Ref(M), Ref(p), Xc, Ref(DefaultOrthogonalBasis()))
+    # X = get_vector.(Ref(M), Ref(p), Xc, Ref(DefaultOrthogonalBasis()))
+    X = hat.(Ref(M), Ref(p), Xc)
     points = exp.(Ref(M), Ref(p), X)
```
dehann, line 129: I would avoid using `Ref()` here during development.  I'd recommend we only add that after things are working.  I'm not sure if Ref will work here, especially with the multithreading code.  Don't believe it is necessary to do that optimization now.

src/ParametricUtils.jl
```diff
@@ -195,6 +220,27 @@ function calcFactorMahalanobisDict(fg)
...
 end
 
+# new experimental Manifold ProductRepr version
+function _totalCost(M, cfvec::Vector{<:CalcFactorMahalanobis}, labeldict, points)
+  #
+  obj = 0
+
+  for cfp in cfvec
+
+      varOrder = getindex.(Ref(labeldict), cfp.varOrder)
```
dehann, line 230: again, julia is already pass by reference, so adding a second tier Ref over a dict ref seems weird to me

test/testSphereMani.jl
```diff
@@ -0,0 +1,46 @@
...
+
+v1 = addVariable!(fg, :x1, Sphere2)
+mf = ManifoldFactor(Sphere(2), MvNormal([0.1, 0.2], [0.05,0.05]))
+f = addFactor!(fg, [:x0, :x1], mf)
+
+##
+# Debugging Sphere error
+# DimensionMismatch("tried to assign 200 elements to 300 destinations")
+smtasks = Task[]
+solveTree!(fg; smtasks, verbose=true, recordcliqs=ls(fg))
```
dehann, line 43: as above, not sure if this is going to work with current KDE.  For the transition to Manifolds.jl, Id suggest we just focus on the types we have for now.  Pose2 Point2 Bearing Range, etc.

examples/dev/testManiFactors.jl
Affie: That sneaked in somehow, File is just a sandbox that can become tests later.

examples/dev/testManiFactors.jl
Affie: It's a good way to catch dimension issues without taking the full step to have matrix or `ProductRepr` for a point.

Affie: But, yes, I changed it to `@test_broken`

src/Factors/GenericFunctions.jl
Affie: It's not an optimization, only for the broadcasting. I can use other containers such as `Tuple`, but prefer `Ref`.  I just find broadcasted code easier to follow and maintain than for loops.

src/ParametricUtils.jl
Affie: Again for broadcasting.

src/ApproxConv.jl
```diff
@@ -168,6 +174,20 @@ function calcVariableCovarianceBasic(ptsArr::Vector{Vector{Float64}})
...
+  Σ = mean(Xcs .* adjoint.(Xcs))
+  @debug "calcVariableCovarianceBasic" μ
+  @debug "calcVariableCovarianceBasic" Σ
+  # TODO don't know what to do here so keeping as before, #FIXME it will break
+  # a change between this and previous is that a full covariance matrix is returned
+  msst = Σ
+  msst_ = 0 < sum(1e-10 .< msst) ? maximum(msst) : 1.0
+  return msst_
+end
+
```
Affie, line 190: @dehann, have a look please

src/ApproxConv.jl
```diff
@@ -168,6 +174,20 @@ function calcVariableCovarianceBasic(ptsArr::Vector{Vector{Float64}})
   return msst_
 end
 
+function calcVariableCovarianceBasic(M::AbstractManifold, ptsArr::Vector{P}) where P
+  #TODO double check the maths,. it looks like its working at least for groups
+  μ = mean(M, ptsArr)
+  Xcs = vee.(Ref(M), Ref(μ), log.(Ref(M), Ref(μ), ptsArr))
+  Σ = mean(Xcs .* adjoint.(Xcs))
```
Affie, line 181:
```suggestion
  Σ = mean(Xcs .* transpose.(Xcs))
```

src/ApproxConv.jl
```diff
@@ -168,6 +174,20 @@ function calcVariableCovarianceBasic(ptsArr::Vector{Vector{Float64}})
...
+  @debug "calcVariableCovarianceBasic" μ
+  @debug "calcVariableCovarianceBasic" Σ
+  # TODO don't know what to do here so keeping as before, #FIXME it will break
+  # a change between this and previous is that a full covariance matrix is returned
+  msst = Σ
```
Affie, line 186: Should this perhaps be the trace to give a single variance:
```suggestion
  msst = tr(Σ)
```

src/ApproxConv.jl
dehann: msst_ is probably part of initialization when trivial case errors occur in this calculation.  I was expecting the first mean for mu, the coordinates on `Xc` look right (around mu).  The adjoint is somewhat unfamiliar to me in this context, must just make sure that the communativity is right, e.g. hypothetically:
```
# good
aRc = adjoint(bRa)*bRc

# vs bad
bRa*adjoint(bRc)
```
since the latter would give incorrect answers.  I'm not sure and would need to dig more to learn more.

src/ApproxConv.jl
dehann: um, probably not that important here since the next line just checks if there are covariances that can work and then scale all according the maximum covariance.  Trace would also work and just scale all by the sum rather than the maximum

dehann:
HI @Affie , the error on serialization has to do with the new ccw.manifolds field:


The issue is that during unpackFactor, we need to construct a ccw without yet knowing what the manifolds are from variables to which it will be attached, 
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/blob/54c128944f02193dfbbb685eb439737be8909d78/src/services/Serialization.jl#L357

see the `DFGVariable[]` argument here:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/9b0c5fc9d0a8d0bfeba8a184a9d241b464d8b2f6/src/DispatchPackedConversions.jl#L55

which creates the problem on the new line of code here (getManifolds on an empty array):
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/9718395b2454e4aa6bd66f7e2c4830bef71b20df/src/FactorGraph.jl#L659

dehann:
ill look more tomorrow

Affie:
About serialization issue of new manifolds field.

Yes, I know it's caused by serialization. I just didn't know how to fix it. 
After yesterday's chat I decided to change it to a vector of variable types and use `getManifold`, ie `Vector{DataType}`

And skip it in the specific test for now.

src/ApproxConv.jl
Affie: Xcs is in coordinates, and transpose is just used to build the matrix to be used for the covariance

src/ApproxConv.jl
Affie: ok

src/ApproxConv.jl
Affie: See #1280

PR #1303, opened 2021-07-09
ccw points Any to fix mixed manifolds bug

Affie:
`Vector{Any}` in `ccw` is the easiest way I can think of to fix this issue. 
I was thinking we get everything running again and then find a better solution 

The tests keep on failing on what looks like unrated numerical tests.


dehann:
Just a note, this was later replaced by a NamedTuple to preserve type stable behaviour.

PR #1307, opened 2021-07-12
fix null and partial tests

dehann:


dehann:
okay, for the first time in a couple of days the IIF tests are passing again. (cc @Affie )

PR #1310, opened 2021-07-16
Use getManifold on factors

Affie:
Hi @dehann, I just want to make sure `getManifold` on factors is the intended way to go

### DIFF
--- a/src/Factors/GenericFunctions.jl
+++ b/src/Factors/GenericFunctions.jl
@@ -73,6 +73,8 @@ struct ManifoldFactor{M <: AbstractManifold, T <: SamplableBelief} <: AbstractMa
     Z::T
 end
 
+DFG.getManifold(f::ManifoldFactor) = f.M
+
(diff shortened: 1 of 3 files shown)


dehann (approved):
HI @Affie , yeah I think `getManifold(fcttype)` is the way to go.  For transition or "easier" cases we can use `ManifoldFactor/Prior` but in general we will probably require the user to implement:
```julia
IIF.getManifold(::InstanceType{MyAwesomeFactor}) = Euclidean(7)
```

since resolving the Manifold is better resolved with dispatch rather than storing the manifold each time as a field in the factor.  The dispatch should be plenty sufficient.

dehann:
Hi @Affie , just also note that I expect the tests to pass, so just make sure everything is fixed before merging this.

Affie:
The tests were broken before this by AMP in https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/107/files#r671323810

Affie:
I see #1309 should fix this.

PR #1311, opened 2021-07-17
Update SE2 manifold test with test_broken

Affie:
@dehann, I added this that was working prevoiusly. Perhaps you can look at it as part of your partials upgrade as it broke in the last week as far as I can tell.


dehann:
so the tests are passing by me, it looks like untested code is failing.  So lets keep adding tests to get the coverage up and make sure it gets debugged with code changes.

PR #1319, opened 2021-07-23
fix solveTree SE2

dehann:
needs JuliaRobotics/ApproxManifoldProducts.jl#120

cc @Affie


dehann:
tests should pass once AMP v0.4.9 is available

PR #1324, opened 2021-07-30
Fix approxDeconv for manifold

Affie:
IIF passes locally and RoME shows improvement. So I'm happy to merge.

### DIFF
--- a/src/DeconvUtils.jl
+++ b/src/DeconvUtils.jl
@@ -71,10 +71,8 @@ function approxDeconv(fcto::DFGFactor,
   # only doing the current active hypo
   @assert activehypo[2][1] == 1 "deconv was expecting hypothesis nr == (1, 1:d)"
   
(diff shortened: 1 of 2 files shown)


src/DeconvUtils.jl
```diff
@@ -71,10 +71,8 @@ function approxDeconv(fcto::DFGFactor,
   # only doing the current active hypo
   @assert activehypo[2][1] == 1 "deconv was expecting hypothesis nr == (1, 1:d)"
   
-  lent = length(makeTarget(1))
-  islen1 = length(makeTarget(1)) == 1
+  islen1 = zDim == 1
```
dehann, line 74: This looks good to me, on the basis that:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/f0fd6f05c57e53c6278aa6904f18a88841ea08f1/src/FactorGraph.jl#L597-L614

PR #1328, opened 2021-08-01
Adding getFactorOperationalMemoryType for NoSolverParams

GearsAD:
This is needed to support getFactor() for DFG types that use NoSolverParams.

### DIFF
--- a/src/IncrementalInference.jl
+++ b/src/IncrementalInference.jl
@@ -410,7 +410,7 @@ const InMemDFGType = DFG.LightDFG{SolverParams}
 import DistributedFactorGraphs: getFactorOperationalMemoryType
 
 getFactorOperationalMemoryType(dfg::SolverParams) = CommonConvWrapper
-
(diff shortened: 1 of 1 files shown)


dehann (approved):
sure, think that should work.  If this works for you then go with it for now until we find new issues.  

Just for info's sake, what I'd like to do with `SolverParams` is refactor to easier-to-follow defaults via [Parameters.jl](https://github.com/mauro3/Parameters.jl).  We [already have defaults](https://github.com/JuliaRobotics/IncrementalInference.jl/blob/e4ce8709bf992153df8f42fe843997e2dae2f5e9/src/FactorGraphTypes.jl#L70), but the code is just a bit messy.  This way CCW can be built against defaults if no parameters are given by the user.  Thus far in IIF/RoME/Caesar solves, there has always been a valid `SolverParams` object.  Haven't tried myself with `NoSolverParams` yet. 

I'm trying to remember off hand if there is strong relationship between CCW and SolverParams, but not immediately remembering anything.

PR #1334, opened 2021-08-03
better organize factor ops memory as entities

dehann:
fyi @Affie


src/services/CalcFactor.jl
```diff
@@ -0,0 +1,186 @@
...
+
+DevNotes
+- Use in place operations where possible and remember `measurement` is a `::Tuple`.
+- TODO only works on `.threadid()==1` at present, see #1094
+- Also see, JuliaRobotics/RoME.jl#465
+"""
+function sampleFactor(cf::CalcFactor{<:AbstractFactor}, 
+                      N::Int=1  )
+  #
+  getSample(cf, N)
```
dehann, line 25: Hi @Affie, here is where getSample gets called from IIF, not sure if there are any other `rand` calls still floating around inside IIF -- they should all come here instead.

src/services/CalcFactor.jl
```diff
@@ -0,0 +1,186 @@
...
+  _, _dfgfct = _buildGraphByFactorAndTypes!(fct, varTypes, pts, dfg=tfg, _blockRecursion=_blockRecursion)
+  
+  # get a fresh measurement if needed
+  _measurement = if length(measurement) != 0
+    measurement
+  else
+    # now use the CommonConvWrapper object in `_dfgfct`
+    ccw = IIF._getCCW(_dfgfct)
+    cfo = CalcFactor(ccw)
+    getSample(cfo, 1)
```
dehann, line 104: and here

PR #1340, opened 2021-08-06
file cleanup and standarization, towards propagateBelief

dehann:


dehann:
going to merge, then revert the compat for Manifolds.jl v0.6 back to v0.5 on master.  From there merge into #1339 and resolve compat for v0.6 there

PR #1357, opened 2021-08-18
rand as coordinates

Affie:
twig that fits with https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/151

### DIFF
--- a/src/SolverUtilities.jl
+++ b/src/SolverUtilities.jl
@@ -20,8 +20,9 @@ DevNotes:
 
 See also: [`getSample`](@ref), [`rand`]
 """
-randToPoints(distr::SamplableBelief, N::Int=1) = [rand(distr,1)[:] for i in 1:N]
-randToPoints(distr::ManifoldKernelDensity, N::Int=1) = rand(distr,N)
+randToPoints(distr, N::Int=1) = [rand(distr,1)[:] for i in 1:N]
+# randToPoints(distr::SamplableBelief, N::Int=1) = [rand(distr,1)[:] for i in 1:N]
+# randToPoints(distr::ManifoldKernelDensity, N::Int=1) = rand(distr,N)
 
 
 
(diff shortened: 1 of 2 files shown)


src/SolverUtilities.jl
```diff
@@ -20,8 +20,9 @@ DevNotes:
 
 See also: [`getSample`](@ref), [`rand`]
 """
-randToPoints(distr::SamplableBelief, N::Int=1) = [rand(distr,1)[:] for i in 1:N]
-randToPoints(distr::ManifoldKernelDensity, N::Int=1) = rand(distr,N)
+randToPoints(distr, N::Int=1) = [rand(distr,1)[:] for i in 1:N]
```
dehann, line 23: Hi, hold up.  We have not changed the behaviour of `rand(::MKD, N::Int=1)`?

I have it that we are going to implement `sample(mkd, e0=identity_element(mkd); as_tangent=true)` for both `::MKD` and `::Distribution`, and default for `sample` will be to return a vector of tangent-vectors.  I'm thinking `rand(::MKD)` will still produce manifold points.

Affie: Hi, yes I know. It is the easiest way to fix a bug and I thought it could be a step in the `sample` update direction. The rand update in AMP is linked above. I was thinking to at least have the test running and then we can do the update and make sure the test remains ok.
We should probably add a similar test to IIF

dehann: i think this PR currently breaks the IIF tests?  So this PR must then either update everything associated with rand, sample, randToPoints -- or find a deprecation path?

Affie: I tested the matching PR in AMP, IIF and RoME(already merged in master as it is an upstream bug) locally.
I'm not opposed to not merging this and fixing it in one go with the new tangent sample

PR #1359, opened 2021-08-20
`sampleTangent` and `samplePoint` 

Affie:
In place of https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/152

from https://github.com/JuliaRobotics/RoME.jl/issues/465

getSample for factors will then use this pattern:
```julia
function getSample(cf::CalcFactor{<:Pose2Pose2}, N::Int=1) 
  
  M = getManifold(Pose2)
  ϵ = getPointIdentity(Pose2)

  X = [sampleTangent(M, cf.factor.z, ϵ) for _=1:N] 
  return (X, )
end
```

### DIFF
--- a/src/Factors/GenericFunctions.jl
+++ b/src/Factors/GenericFunctions.jl
@@ -63,7 +63,8 @@ function getSample(cf::CalcFactor{<:ManifoldFactor}, N::Int=1)
     (ret, )
 end
 
-function (cf::CalcFactor{<:ManifoldFactor{<:AbstractGroupManifold}})(Xc, p, q)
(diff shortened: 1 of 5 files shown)


dehann:
HI @Affie , that pattern looks good to me thanks -- lets go with `sampleTangent`.  Future versions can resolve wheter the `N::Int` part as internal to IIF or user provided feature.

dehann (approved):

Affie:
I'm merging as it is passing tests. will make another pr for updates to factors to use it as it's bigger than anticipated.

PR #1368, opened 2021-08-24
Try getManifold fall back to getManifoldPartial to get the manifold used in partial priors

Affie:
So far only tested with Partial Pose3 https://github.com/JuliaRobotics/RoME.jl/pull/495 but numerics fail

Updated PartialDim2{T} <: AbstractPrior to use getManifold method, while PartialPrior falls back to using getManifoldPartial

Just note that I do not know what to do with factors like `PartialPrior` that do not have a manifold assigned. I think it would be best to have all factors assigned a Manifold with Euclidean/TranslationGroup for the current linear ones.


Affie:
I have no idea why this is failing yet. Sphere2 tests still works locally

PR #1380, opened 2021-09-09
Fix and test manifold approxDeconv

### LINKED ISSUE #1374: _solveLambdaNumeric for AbstractManifoldMinimize does not work for deconv

Affie:
close #1374

Note: only for measurements as tangent vectors


dehann:
> Note: only for measurements as tangent vectors

thanks @Affie , i added a link to requirements

PR #1389, opened 2021-09-15
Update to  RealCircularGroup

### LINKED ISSUE #1365: Upgrade to RealCircleGroup
Circular is not currently working and tests are skipped until we upgrade to `CircleGroup`.

@dehann, we need https://github.com/JuliaManifolds/Manifolds.jl/issues/415 to continue.

Affie:
I changed to RealCircularGroup as it was the closest to what we had and it works.
close #1365

This PR also looks at #150, but in IIF to use <:InferenceVariable

### DIFF
--- a/src/ManifoldSampling.jl
+++ b/src/ManifoldSampling.jl
@@ -28,7 +28,7 @@ function sampleTangent(M::AbstractManifold, z::Distribution, p, basis::AbstractB
 end
 
 function sampleTangent(M::AbstractGroupManifold, z::Distribution, p=identity_element(M)) 
-  return hat(M, p, rand(z))
+  return hat(M, p, rand(z,1)[:]) #TODO find something better than (z,1)[:]
 end
 
 
@@ -40,7 +40,7 @@ Return a random sample point on a manifold from a belief represented by coordina
 Notes
 
 """
-function samplePoint(M::AbstractManifold, sbelief, p, basis, retraction_method::AbstractRetractionMethod=ExponentialRetraction())
+function samplePoint(M::AbstractManifold, sbelief, p, basis::AbstractBasis, retraction_method::AbstractRetractionMethod=ExponentialRetraction())
   X = sampleTangent(M, sbelief, p, basis)
   return retract(M, p, X, retraction_method)
 end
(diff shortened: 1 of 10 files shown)


src/Variables/DefaultVariables.jl
```diff
@@ -5,7 +5,8 @@ export Circular, Circle
 
 
 Base.convert(::Type{<:Tuple}, ::InstanceType{Manifolds.Euclidean{Tuple{N}, ℝ}} ) where N = tuple([:Euclid for i in 1:N]...)
-Base.convert(::Type{<:Tuple}, ::InstanceType{Manifolds.Circle{ℝ}})  = (:Circular,)
+Base.convert(::Type{<:Tuple}, ::InstanceType{Manifolds.Circle{ℝ}})  = error("#FIXME")#(:Circular,)
+Base.convert(::Type{<:Tuple}, ::InstanceType{Manifolds.RealCircleGroup})  = (:Circular,)
```
dehann, line 9: yeah, I think we should just ditch the old tuples from IIF v0.25 entirely.  Doesn't block this PR though, must just remember to do before release.

src/services/EvalFactor.jl
```diff
@@ -411,7 +412,8 @@ function evalPotentialSpecific( Xi::AbstractVector{<:DFGVariable},
 
   # view on elements marked for nullhypo
   addEntrNH = view(addEntr, nhmask)
-  spreadDist = spreadNH*sqrt(calcCovarianceBasic(mani, addEntr))
+  # spreadDist = spreadNH*sqrt(calcCovarianceBasic(mani, addEntr))
+  spreadDist = spreadNH*calcStdBasicSpread(getVariableType(Xi[sfidx]), addEntr)
```
dehann, line 416: thanks for fixing this and other use in IIF.

dehann (approved):
Code read looks good to me, I just restarted the test.  I'd like to make sure the IIF tests pass on this PR before we merge.

dehann:
oops, this file keeps failing again and again:
`test/testExpXstroke.jl`

we must get this fixed first before we can merge

src/ManifoldSampling.jl
```diff
@@ -40,15 +40,22 @@ Return a random sample point on a manifold from a belief represented by coordina
...
+function samplePoint(M::AbstractManifold, sbelief, p, basis::AbstractBasis, retraction_method::AbstractRetractionMethod=ExponentialRetraction())
   X = sampleTangent(M, sbelief, p, basis)
   return retract(M, p, X, retraction_method)
 end
 function samplePoint(M::AbstractGroupManifold, sbelief, p=identity_element(M), retraction_method::AbstractRetractionMethod=ExponentialRetraction())
   X = sampleTangent(M, sbelief, p)
   return retract(M, p, X, retraction_method)
 end
+
+function samplePoint(M::AbstractGroupManifold, sbelief::ManifoldKernelDensity, p=mean(sbelief), retraction_method::AbstractRetractionMethod=ExponentialRetraction())
```
Affie, line 52: Just a breadcrumb, new numerical failure might stem from this change. I can’t test now though.

Affie: Yes. This is the issue. MKD is not calculated around the mean but identity element, so should use `identity_element`
I’ll fix if i get proper signal.

src/services/EvalFactor.jl
Affie: Maybe a better verb-noun is `calcSpreadBasic`

Issue #1424, opened 2021-10-17
Replace PackedFactor convert functions with specific function. Two chicken and egg problems, unpack DFGFactor


dehann:
We should NOT use driving assumption that `Base.convert` API is right way to hydrate DFGFactor (which currently internally rebuilds operational data like CCW).  
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/88c0cb491c5ec16f74078f063d3a3007da4296ae/src/DispatchPackedConversions.jl#L33-L55

By assuming empty array `DFGVariable[]` we create two chicken egg problems:
- recursion on delete / add new factor but requiring a rebuild metadata in both:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/88c0cb491c5ec16f74078f063d3a3007da4296ae/src/DispatchPackedConversions.jl#L87
- Solver needs needs to then replace the skeleton CCW metadata type when the variable types are more easily known:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/88c0cb491c5ec16f74078f063d3a3007da4296ae/src/FactorGraph.jl#L681

The assumption that we can hydrate a factor in a vacuum is false, we do need the connected DFGVariable type information.  DFGVariable can however be hydrated in a vacuum.

GearsAD:
Fix this when fixing https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/590

dehann:
we need to maintain the list of exceptions to DFG driver symmetries.  In this case, the exception is:
- rebuildFactorMetadata / OperationalMem is not required in all DFG drivers.
  - CDDFG supporting direct convolution and product operations is Tier 2 requirement at best (symmetry is important, but FactorOperationalMemory is needed and only exists in LightDFG).

sam-globus:
Yep, 

LightDFG is our DFG for solving - it's optimized for this.

CGDFG and FileDFG (and ZmqDFG/GraphQLDFG, etc.) are all designed to get data from somewhere and put it into a LightDFG, and then transport solved data back into their respective destination. Those aren't built for solving/etc, to the point that _ideally_ we don't rebuild the factor metadata.

dehann:
Added to High Level Req Wiki: https://github.com/JuliaRobotics/Caesar.jl/wiki/High-Level-Requirements

PR #1425, opened 2021-10-17
Fix for CGDFG during `rebuildFactorMetadata`

GearsAD:


dehann:
xref 
- #1424 
- JuliaRobotics/DistributedFactorGraphs.jl#809

PR #1438, opened 2021-10-23
Bson testTreeSaveLoad for julia 1.7

Affie:


dehann:
probably easy to switch to BSON to store trees instead.

PR #1439, opened 2021-11-01
Default getSample methods for group manifolds

Affie:


dehann (approved):

dehann:
From the call today, two comments came up:
- Check `hasproperty` for `.Z` and give good warning for user to easily understand
- Lets standardize to only field `.Z` as the automated field for both non-parametric and parametric,

xref #1441

Added to requirements wiki:

Affie:
@dehann, comments implemented.

Affie:
I don't think this is a breaking change, should it not be in v0.25.6?

src/ManifoldSampling.jl
```diff
@@ -65,5 +65,46 @@ function samplePoint(distr::SamplableBelief)
...
+- See the Custom Factors section in the Caesar.jl documentation for more examples and details.
+- Also see issue https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1441
+
+See also: [`getMeasurementParametric`](@ref)
+"""
+function getSample end
+
+function getSample(cf::CalcFactor{<:AbstractPrior}) 
+  M = getManifold(cf.factor)
+  if hasfield(typeof(cf.factor), :Z)
```
dehann, line 92: perhaps we should standardize with API
```julia
getMeasurement
```
to return the field `.Z` etc.  This can then also be used the existing function `getMeasurementParametric` which extracts the mean and covariance for parametric solves?

dehann (approved):

Issue #1393, opened 2021-09-26
solve on larger graph empty collection error


dehann:
```julia
using RoME
# also had Distributed.addprocs(6)

fg = generateCanonicalFG_Beehive!(40, locality=2)
getSolverParams(fg).useMsgLikelihoods = true

# more testing needed to see if stochastic bug for smallish graphs

solveTree!(fg)


ArgumentError: reducing over an empty collection is not allowed
        Stacktrace:
          [1] _empty_reduce_error
            @ ./reduce.jl:299
          [2] reduce_empty
            @ ./reduce.jl:309
          [3] mapreduce_empty
            @ ./reduce.jl:343
          [4] reduce_empty
            @ ./reduce.jl:329
          [5] reduce_empty_iter
            @ ./reduce.jl:355 [inlined]
          [6] mapreduce_empty_iter
            @ ./reduce.jl:351
          [7] _mapreduce
            @ ./reduce.jl:400 [inlined]
          [8] _mapreduce_dim
            @ ./reducedim.jl:318 [inlined]
          [9] #mapreduce#672
            @ ./reducedim.jl:310 [inlined]
         [10] mapreduce
            @ ./reducedim.jl:310 [inlined]
         [11] #_maximum#690
            @ ./reducedim.jl:878 [inlined]
         [12] _maximum
            @ ./reducedim.jl:878 [inlined]
         [13] #_maximum#689
            @ ./reducedim.jl:877 [inlined]
         [14] _maximum
            @ ./reducedim.jl:877 [inlined]
         [15] #maximum#687
            @ ./reducedim.jl:873 [inlined]
         [16] maximum
            @ ./reducedim.jl:873 [inlined]
         [17] #propagateBelief#330
            @ ~/.julia/dev/IncrementalInference/src/GraphProductOperations.jl:48
         [18] #localProduct#336
            @ ~/.julia/dev/IncrementalInference/src/GraphProductOperations.jl:108
         [19] #localProductAndUpdate!#340
            @ ~/.julia/dev/IncrementalInference/src/GraphProductOperations.jl:135
         [20] #106
            @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Distributed/src/process_messages.jl:278
         [21] run_work_thunk
            @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Distributed/src/process_messages.jl:63
         [22] macro expansion
[...]

PR #1463, opened 2021-12-05
refactor getDimension, standardization, sampling, Flux, fix

dehann:


dehann:
partially or completely fixes #1393

PR #1472, opened 2022-01-17
Towards common location for Serializing Distributions/Density (contracts)

dehann:
Needs DFG v0.18


dehann:
Merging against DFG main|master (upstream Dev CI), although DFG v0.18 has not yet been tagged

PR #1479, opened 2022-02-03
Standardise to LocalDFG in IIF

### LINKED ISSUE #1478: Update local instances to use name LocalDFG
With DFG v0.18.0, probably a good idea to standardize the names better.  Remove all references to LightDFG.  See:
- JuliaRobotics/DistributedFactorGraphs.jl#826
- JuliaRobotics/DistributedFactorGraphs.jl#844

Affie:
Close #1478


src/ExportAPI.jl
```diff
@@ -3,9 +3,10 @@
 
 # DFG SpecialDefinitions
 export AbstractDFG,
-  InMemDFGType,
   getSolverParams,
   LightDFG,
+  GraphsDFG,
```
dehann, line 8: should probably move export of LightDFg to Deprecated.jl

dehann (approved):

src/ExportAPI.jl
Affie: Ok, sounds good, will update

PR #1484, opened 2022-02-10
Remove inner constructor PackedGenericMarginal

### LINKED ISSUE #1483: duplicate loading of PackedGenericMarginal
```julia
WARNING: Method definition (::Type{IncrementalInference.PackedGenericMarginal})() in module IncrementalInference at /home/dehann/.julia/dev/IncrementalInference/src/Factors/GenericMarginal.jl:21 overwritten at util.jl:478.
  ** incremental compilation may be fatally broken for this module **
```

Affie:
close #1483


dehann (approved):
Ah, yeah thanks!  we should remove the inner constructors that aren't explicitly necessary.  I've been pulling them out too.

PR #1529, opened 2022-06-01
Split tests into two groups

### LINKED ISSUE #1528: Consider splitting CI tests into two parts
I propose we split the CI tests into two parts to cut the time for tests and enable re-running numerical "unstable" tests separately.
For example:
1. Functional and Basic tests
2. Numerical tests cases

Affie:
Pattern from Manifolds.jl
close #1528


Affie:
@dehann, here is a draft pr for splitting the tests, it seems to work well with faster tests and we can now only restart the failed part.

dehann:
sounds good!

PR #1535, opened 2022-06-19
more standard initVariable!

dehann:
still need to update all the other uses of `initManual!` elsewhere in the code base.  Best to do that in a separate PR before v0.29 is tagged:
- #1536


test/testSpecialEuclidean2Mani.jl
```diff
@@ -105,6 +106,31 @@ pbel_ = approxConvBelief(fg, :x0f1, :x0)
 ##
 end
 
+
+@testset "test initVariableManual! with Vector of Tuple inputs" begin
+##
+
+fg = initfg()
+
+pts = [(randn(2), _Rot.RotMatrix2(randn()).mat) for _ in 1:50]
```
dehann, line 115: FYI @jim-hill-r

src/services/GraphInit.jl
```diff
@@ -227,46 +227,62 @@ function initManual!( variable::DFGVariable,
...
+                        bw=nothing ) where {P}
   #
-  var = getVariable(dfg, sym)
-  M = getManifold(var)
-  pp = manikde!(M, pts, bw=bw)
-  initManual!(var,pp, solveKey)
+  # specializations to support generic case of Tuple rather than ProductRepr or ArrayPartition inputs
+  # TODO ArrayPartition inputs
+  _prodrepr(pt) = pt
+  _prodrepr(pt::Tuple) = Manifolds.ProductRepr(pt...)
```
dehann, line 267: @jim-hill-r , so the problem was we are actually using `Manifolds.ProductRepr` to represent on-manifold points.  This is slowly converting to a more standardized format with wider JL community as `ArrayPartition`.  To simplify all that, I making the assumption that if we are reserving `Tuple[...]` of points as a API shorthand which must to automatically converted to `ProductRepr[...]`.  We can resolve this once the `ArrayPartition` change has run it's course:
- #1537 

cc @Affie

test/testSpecialEuclidean2Mani.jl
```diff
@@ -105,6 +106,31 @@ pbel_ = approxConvBelief(fg, :x0f1, :x0)
...
+
+addVariable!(fg, :x0, SpecialEuclidean2)
+initVariable!(getVariable(fg, :x0), pts)
```
dehann, line 118: oh, and here is the init with vector of tuple

PR #1552, opened 2022-07-08
Change to support ArrayPartition by default

Affie:


Affie:
It passes on upstream dev, so going to merge to keep things going

dehann:
looks good thanks

PR #1568, opened 2022-07-18
solveGraphParametric returns points on manifold

Affie:
Make sure it is consistent everywhere. It was coordinates in updateGraphParametric and points in tree solve.
Should also add tests.

- [x] fix `solveConditionalsParametric`, perhaps consolidate with `solveGraphParametric` see #1584
- [x] add tests for autoinit 
-  RoME tests will need to be updated

### DIFF
--- a/src/ParametricUtils.jl
+++ b/src/ParametricUtils.jl
@@ -805,9 +805,11 @@ function autoinitParametric!(dfg::AbstractDFG,
 
     # updateVariableSolverData!(dfg, xi, solveKey, true; warn_if_absent=false)    
     # updateVariableSolverData!(dfg, xi.label, getSolverData(xi, solveKey), :graphinit, true, Symbol[]; warn_if_absent=false)
+  else
+    result = nothing
   end
 
-  return isInitialized(xi, solveKey)
+  return result#isInitialized(xi, solveKey)
 end
 
 ## ================================================================================================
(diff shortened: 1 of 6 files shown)


src/ParametricUtils.jl
```diff
@@ -653,41 +654,6 @@ function solveConditionalsParametric(fg::AbstractDFG,
...
-- Consolidate around `ManifoldsBase.AbstractManifold` instead, with possible wrapper-type solution for `Optim.Manifold`
-"""
-struct MixedCircular <: Optim.Manifold
-  isCircular::BitArray
-end
-
-function MixedCircular(fg::AbstractDFG, varIds::Vector{Symbol})
-  circMask = Bool[]
-  for k = varIds
-    append!(circMask, convert(Tuple, getManifold(getVariableType(fg, k))) .== :Circular)
```
dehann, line 674: yeah, think it will be good to remove this older code.  Assume this has all been replaced by the Product and PowerManifold stuff?

Affie: Yes, and this only worked on the old :circular and :euclidian

Affie:
I suspect something upstream might have broken tests as tests are failing that are unrelated to this PR.

dehann:
Resolved in 
- #1576

PR #1569, opened 2022-07-18
OrderedDict changed variable ordering

Affie:
This passes on DFG master.
Perhaps the correct thing to do here is to manually force a variable ordering for these tests.


dehann (commented):
looks good

Issue #1570, opened 2022-07-18
"Basic 3 door, trinary multihypothesis tests" often fails numerically


Affie:

dehann:
This must be a regression.  It was working well.

Affie:
I checked v0.28 and it was failing intermittently there as well, so think we can safely eliminate the recent `ArrayPartition` and Manifolds v0.8 upgrade.

dehann:
What I mean is the frequency of failures seem much higher to me than I remember v0.26-28 being.  I looked into this a bit and could not immediately see anything wrong.  Perhaps the test is just too strict, or we should change the testing strategy on this file for more amenable outcomes to current performance.

Affie:
From #1580, it looks like this will only be done after #1010

dehann:
i still think this test used to be better.  I looked at it he other day and seems like it's "working" just a bit less often than before.  So there is possibly as soft bug somewhere, but let's keep moving until something more shows up

PR #1572, opened 2022-07-25
v0.30.1-rc1

dehann:
Hi @Affie , did you want to add anything to v0.30.1, or can the next PRs go to v0.30.2?


dehann:
I am worried about 
- #1570 

So will look into that bit later today if can.

dehann:
Caesar tests also seem to be failing more often on IIF master.  I get the sense that some of the recent changes may have upset the multihypo code with silent bug somewhere.

Affie:
It can go into the next, but master broke from something upstream (i think)

Affie:
Technically parametric has a bug in with the return types, but if you are in a hurry for something you can tag.

dehann:
Looks like it has something to do with `getSolverParams(fg).useMsgLikelihoods = true`.  MWE:

```julia
using IncrementalInference

getSolverParams(fg).useMsgLikelihoods = true

solveGraph!(fg)
```

dehann:
useMsgLikelihoods uses `approxDeconv`, and I wonder if all that code got updated for Tangents vs Points in the new ArrayPartition work.

PR #1575, opened 2022-07-26
Too many getPointIdentity, #1574

dehann:
close:
- #1574

PR #1576, opened 2022-07-26
Fix samplTangent MKD coords, and test

### LINKED ISSUE #1574: Regression, new bug, related to useMsgLikelihoods
Looks like it has something to do with `getSolverParams(fg).useMsgLikelihoods = true`.  MWE:

```julia
using IncrementalInference

getSolverParams(fg).useMsgLikelihoods = true

solveGraph!(fg)
```

Code completes fine if `useMsgLikelihoods=false`

_Originally posted by @dehann in https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1572#issuecomment-1194719417_

Affie:
Pick some changes from #1575
Only changes I understood and knew would not break anything else. 
See .. diff: https://github.com/JuliaRobotics/IncrementalInference.jl/compare/22Q3/bug/1574_test..22Q3/bugfix/twig_1574

close #1574

### DIFF
--- a/src/ManifoldSampling.jl
+++ b/src/ManifoldSampling.jl
@@ -11,10 +11,11 @@ function sampleTangent end
 
 # Sampling MKD
 function sampleTangent(M::AbstractDecoratorManifold, x::ManifoldKernelDensity, p=mean(x))
-    # get legacy matrix of coordinates and selected labels
-    coords, lbls = sample(x.belief,1)
-    X = hat(x.manifold, p, coords)
-    return X
+  # get legacy matrix of coordinates and selected labels
+  #TODO make sure that when `sample` is replaced in MKD, coordinates is a vector
+  coords, lbls = sample(x.belief,1)
+  X = hat(x.manifold, p, coords[:])
+  return X
 end
 
 function sampleTangent(x::ManifoldKernelDensity, p=mean(x)) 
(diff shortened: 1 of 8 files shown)


src/ManifoldSampling.jl
```diff
@@ -11,10 +11,11 @@ function sampleTangent end
...
 # Sampling MKD
 function sampleTangent(M::AbstractDecoratorManifold, x::ManifoldKernelDensity, p=mean(x))
-    # get legacy matrix of coordinates and selected labels
-    coords, lbls = sample(x.belief,1)
-    X = hat(x.manifold, p, coords)
-    return X
+  # get legacy matrix of coordinates and selected labels
+  #TODO make sure that when `sample` is replaced in MKD, coordinates is a vector
+  coords, lbls = sample(x.belief,1)
+  X = hat(x.manifold, p, coords[:])
```
Affie, line 17: `coords[:]` is the only change needed to fix the broken master.

Affie: The current sampling state is:
- sampling from beliefs (eg `sample` and `rand`) are in coordinates.
- sampleTangent is on the tangent at point `p` or the identity element for groups.

Affie:
I think we should merge this and then look at the rest of #1575 again.

dehann (approved):

src/ManifoldSampling.jl
dehann: See JuliaRobotics/Caesar.jl#895

PR #1585, opened 2022-08-05
Test with a different AffineSimplexer

Affie:


Affie:
This will become obsolete when we use Manopt

PR #1588, opened 2022-08-10
future consolidation of parametric factor eval

dehann:


Affie:
Would you mind if we keep it in the same file, it shares 90% with `solveGraphParametric` and I wanted to see what I can consolidate there first. I spend a few hours on it but left it as I thought it was a low priority. I was thinking of doing it with the tree solve updates.

And then about moving code in general: It makes tracking changes impossible if code is moved with changes. 
In the past, I've tried to do a separate PR/commit for moving code with no changes and named it as such. 
I use git blame a lot to see why something changed if I can't remember it or want to find out why someone else made a change.

dehann:
okay, no problem.

PR #1589, opened 2022-08-10
move MaxMixture funcs separate file

dehann:


Affie:
My preference is for only stable code to be moved around to not lose git blame history, But I'm ok with this if you want it.

Affie (approved):

PR #1594, opened 2022-08-11
Improve allocations of get_vector! and exp! and autoinitParametric

Affie:
This has the following improvement:
```julia
@benchmark gsc(Xc) setup=(Xc=randn($N))

# BenchmarkTools.Trial: 10000 samples with 1 evaluation.
#  Range (min … max):  27.226 μs …  12.187 ms  ┊ GC (min … max): 0.00% … 99.35%
#  Time  (median):     28.625 μs               ┊ GC (median):    0.00%
#  Time  (mean ± σ):   32.140 μs ± 168.955 μs  ┊ GC (mean ± σ):  7.40% ±  1.40%

#   ▂▇█▇▆▄▄▃▂▂▁                                                  ▂
#   ████████████▇▆▅▆▄▄▄▄▁▁▃▁▃▁▁▃▆▇▇█▇▇▇▆▆▅▅▅▆▇▇▇▇▇▆▇▆▆▅▃▅▄▃▄▅▆▆▅ █
#   27.2 μs       Histogram: log(frequency) by time      55.4 μs <

#  Memory estimate: 23.08 KiB, allocs estimate: 458.

# with manifolds fix for get_vector! and exp! in ManifoldsExtentions
# and static array iΣ
# BenchmarkTools.Trial: 10000 samples with 1 evaluation.
#  Range (min … max):  11.599 μs …  10.326 ms  ┊ GC (min … max): 0.00% … 99.50%
#  Time  (median):     13.096 μs               ┊ GC (median):    0.00%
#  Time  (mean ± σ):   15.114 μs ± 103.187 μs  ┊ GC (mean ± σ):  6.80% ±  0.99%

#    ▄█▂▃                                                         
#   ▄█████▆▅▄▅▅▄▃▃▃▂▂▂▂▂▂▂▂▂▁▂▁▂▁▁▁▁▁▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
#   11.6 μs         Histogram: frequency by time         32.7 μs <

#  Memory estimate: 12.14 KiB, allocs estimate: 203.
```


dehann:
We'll get the allocation numbers down.  At least now all the effort is aligned with the community around Manifolds.jl, so i suspect the commons induction will eventually become quite dramatic.

Affie:
@dehann, just note I moved autoInitParametric back to parametricUtils as a merge conflict caused a duplication of the function and I had to manually sort it out.

Affie:
I ran out of time to clean this up. Will continue soon.

dehann:
Hi @Affie , I've made a new CI test sequence for IIF which is much friendlier on known numerical failures.  Nothing to do with this PR, but thought you'd perhaps like to know and wanted to merge that in here for better testing.  Will merge soon:
- #1606

Affie:
Sure, this will not pass yet as I need https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/pull/891

src/ParametricUtils.jl
```diff
@@ -774,6 +854,81 @@ function createMvNormal(v::DFGVariable, key=:parametric)
...
+  autoinitParametric!(dfg, getVariable(dfg,initme); kwargs...)
+end
+
+
+function autoinitParametric!(
+    dfg::AbstractDFG,
+    xi::DFGVariable;
+    solveKey = :parametric,
+    reinit::Bool=false,
+    kwargs...
```
Affie, line 895: @dehann, one of the changes is we can now pass Optim parameters to the init function, for example:
```julia  
  algorithm=Optim.NelderMead,
   algorithmkwargs = (initial_simplex = Optim.AffineSimplexer(0.25, 0.1),)
```

src/ManifoldsExtentions.jl
```diff
@@ -53,16 +77,16 @@ end
 
 #FIXME fix back to SA
 function getPointIdentity(G::SpecialOrthogonal{N},::Type{T}=Float64) where {N,T<:Real}
-  # return SMatrix{N,N, T}(I)
-  return Matrix{T}(I, N, N)
+  return SMatrix{N,N, T}(I)
+  # return Matrix{T}(I, N, N)
```
Affie, line 81: Note to self: "Myself, remember to switch these back"

src/ParametricUtils.jl
dehann: ah cool, okay.  I don't know too much about init presets yet but excited to try and learn.  Sure we've probably been using the default which is bad for us.  Looks like you did some digging and selected AffineSimplexer.  Also get the idea we both like Nelder Mead more than BGFS :-)

Affie: For small problems NM works surprisingly well but not with the default Simplex at 0. For medium BFGS looks really good. Large I still have to figure out.

dehann: Large would be L-BGFS, but might be moot since that may be using sparsity in the same we are doing with the tree.  Cliques should be be small or at worst medium in size (exp complexity in clique dimension).  That's the same reason why BGFS is bad on large dense problems, so my guess is L-BGFS is leveraging sparsity.

dehann:
to help this branch, I suggest first merging:
- #1607

Affie:
Without using StaticArrays, this still gets us to here:

BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  11.940 μs …  10.926 ms  ┊ GC (min … max): 0.00% … 99.52%
 Time  (median):     13.594 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   16.524 μs ± 109.235 μs  ┊ GC (mean ± σ):  6.58% ±  1.00%

  ▅▆█▇▅▅▆▇▅▃▃▂               ▁▁▂▁  ▁     ▁▁▁                   ▂
  ████████████▇▅▅▅▃▃▁▃▁▃▁▁▁▁▃█████████▇▇▇████▇▇▇▇▆▆▅▆▅▅▆▅▃▅▅▇▇ █
  11.9 μs       Histogram: log(frequency) by time      40.3 μs <

 Memory estimate: 13.55 KiB, allocs estimate: 242.

src/ManifoldsExtentions.jl
dehann: did you remember?

src/ParametricUtils.jl
```diff
@@ -700,13 +781,15 @@ Initialize the parametric solver data from a different solution in `fromkey`.
 function initParametricFrom!(fg::AbstractDFG, fromkey::Symbol = :default; parkey::Symbol = :parametric, onepoint=false)
   if onepoint
     for v in getVariables(fg)
+      isInitialized(v, parkey) && continue
```
dehann, line 784: don't think we should do that.  If you init a graph based on other solveKey then that should be absolute.  Not have cases that silently skip since I expect the user is trying to force a particular initialization.

dehann (changes requested):

src/ParametricUtils.jl
```diff
@@ -700,13 +781,15 @@ Initialize the parametric solver data from a different solution in `fromkey`.
...
       fromvnd = getSolverData(v, fromkey)
       dims = getDimension(v)
       getSolverData(v, parkey).val[1] .= fromvnd.val[1]
       getSolverData(v, parkey).bw[1:dims,1:dims] .= LinearAlgebra.I(dims)
     end
   else
     for var in getVariables(fg)
+        isInitialized(var, parkey) && continue
```
dehann, line 792:
```suggestion
```

src/ParametricUtils.jl
dehann:
```suggestion
```

PR #1597, opened 2022-08-13
initAll! for `:parametric`

dehann:


src/ParametricUtils.jl
```diff
@@ -775,47 +775,5 @@ function createMvNormal(v::DFGVariable, key=:parametric)
 end
 
 
-function autoinitParametric!(dfg::AbstractDFG,
```
Affie, line 778: This is still under development and was changed in #1594. I'm also working on it locally as it's not reliable at all. Moving files around makes it a manual process to track changes and manage conflicts which I do not have time for now.

PR #1611, opened 2022-08-24
Remove FactorMetaData

### LINKED ISSUE #1563: Remove FMd (should be replaced by CalcFactor)
~First CPT~ #1562 

This first and then #1562

Affie:
Outstanding: 
- [x] should `fullvariables` be filtered by hypo
- [x] `solvefor` weirdness in DERelative
- [x] update news for breaking changes
- [x] bump minor version

Also, CPT should be removed in the same release to avoid possible issues.

This will close #1563


src/entities/FactorOperationalMemory.jl
```diff
@@ -42,6 +39,12 @@ struct CalcFactor{T <: AbstractFactor, M, P <: Union{<:Tuple,Nothing,AbstractVec
   _allowThreads::Bool
   """ user cache of arbitrary type, overload the [`preambleCache`](@ref) function. NOT YET THREADSAFE """
   cache::C
+
+  ## TODO Consolidation WIP with FactorMetadata
+  # full list of variables connected to the factor
+  fullvariables::Vector{DFGVariable}
```
dehann, line 45: ah dynamic types.  Let's do Vector just to get rid of FMd and the come back for `{T<:Tuple}`?

src/entities/FactorOperationalMemory.jl
```diff
@@ -181,6 +153,8 @@ mutable struct CommonConvWrapper{ T<:AbstractFactor,
   _gradients::G
   # type used for cache
   dummyCache::CT
+  #
+  fullvariables::Vector{DFGVariable}
```
dehann, line 157: yeah, cleanup of CCW after FMd and CPT are removed.  That should start to help performance, allocations, dynamic dispatch etc.

src/services/CalcFactor.jl
```diff
@@ -20,44 +20,31 @@ _getDimensionsPartial(fct::DFGFactor) = _getDimensionsPartial(_getCCW(fct))
...
+                    factor = ccwl.usrfnc!,
                     _sampleIdx = 0, 
                     _measCount = length(ccwl.measurement),
                     _legacyMeas = ccwl.measurement, 
                     _legacyParams = ccwl.params, 
                     _allowThreads = true,
-                    cache = ccwl.dummyCache )
+                    cache = ccwl.dummyCache,
+                    fullvariables=ccwl.fullvariables,
+                    solvefor=ccwl.varidx)
```
dehann, line 36: we are also going to have to double check this with the multihypo recipes.  Basically the ground rule is that a CalcFactor object represents one specific hypothesis.  If there are particles with different hypos, then each particle gets a different CalcFactor object.

From that, `.fullvariables` and `solvefor` should only refer to the one active hypothesis this CalcFactor object is working with.

src/services/CalcFactor.jl
```diff
@@ -240,11 +224,11 @@ function CommonConvWrapper( usrfnc::T,
                             res::AbstractVector{<:Real}=zeros(zDim),
                             threadmodel::Type{<:_AbstractThreadModel}=MultiThreaded,
                             inflation::Real=3.0,
-                            vartypes=typeof.(getVariableType.(factormetadata.fullvariables)),
+                            vartypes=DataType[],
```
dehann, line 227: Why not just:
```suggestion
                            vartypes=typeof.(getVariableType.(fullvariables)),
```

src/services/CalcFactor.jl
```diff
@@ -240,11 +224,11 @@ function CommonConvWrapper( usrfnc::T,
...
                             gradients=nothing,
-                            userCache::CT = nothing ) where {T<:AbstractFactor,P,H,CT}
+                            userCache::CT = nothing,
+                            fullvariables=DFGVariable[] ) where {T<:AbstractFactor,P,H,CT}
```
dehann, line 230:
```suggestion
                           ) where {T<:AbstractFactor,P,H,CT}
```

src/services/CalcFactor.jl
dehann: > should fullvariables be filtered by hypo

Yes it should be filtered by hypo

test/testMultihypoFMD.jl
```diff
@@ -16,14 +16,14 @@ end
...
-  @warn "getSample(cf::CalcFactor{<:MyFactor},::Int) does not get hypo sub-selected FMD data"
-  @show DFG.getLabel.(cf.metadata.fullvariables)
+  @warn "getSample(cf::CalcFactor{<:MyFactor},::Int) does not get hypo sub-selected FMD data: $(DFG.getLabel.(cf.fullvariables))" cf.solvefor maxlog=1
   # @assert DFG.getLabel.(fmd_[1].fullvariables) |> length < 3 "this factor is only between two variables"
   return rand(cf.factor.Z, 1)
 end
 
 function (cf::CalcFactor{<:MyFactor})(z, X1, X2)
-  @assert DFG.getLabel.(cf.metadata.fullvariables) |> length < 3 "this factor is only between two variables"
+  length(cf.fullvariables) < 3 && @warn("this factor is only between two variables. solvefor=$(cf.solvefor)", maxlog=1)
```
dehann, line 25:
```suggestion
  @assert length(cf.fullvariables) < 3 "this factor is only between two variables. solvefor=$(cf.solvefor)"
```

This error is important in testing

src/entities/FactorOperationalMemory.jl
Affie: Yes, currently it’s dispatched dynamically in any case and this way it at least saves on compile time.
But the plan is to figure out the best way after the cleanup and not waste time now.

src/services/CalcFactor.jl
Affie: Ok. I figured that much, but couldn’t quite see how it was done / can be done without first removing CPT.

src/services/CalcFactor.jl
Affie: julia> typeof.(getVariableType.(DFGVariable[]))
Any[]

Affie: But it should be done better

test/testMultihypoFMD.jl
Affie: Jip, I fixed it, was just to test with and help me understand.

src/services/CalcFactor.jl
Affie: done

Affie:
We should probably tag a new patch release before merging this into master.

PR #1615, opened 2022-08-26
Remove CPT

### LINKED ISSUE #1562: Remove CPT (should be replaced by CalcFactor)
First doing #1563

Affie:
close #1562

### DIFF
--- a/src/entities/FactorOperationalMemory.jl
+++ b/src/entities/FactorOperationalMemory.jl
@@ -22,73 +22,41 @@ DevNotes
 
 Related 
 
-[`CalcFactorMahalanobis`](@ref), [`CommonConvWrapper`](@ref), [`ConvPerThread`](@ref)
+[`CalcFactorMahalanobis`](@ref), [`CommonConvWrapper`](@ref)
 """
-struct CalcFactor{T <: AbstractFactor, P <: Union{<:Tuple,Nothing,AbstractVector}, X, C}
-  """ the interface compliant user object functor containing the data and logic """
-  factor::T
-  """ what is the sample (particle) id for which the residual is being calculated """
-  _sampleIdx::Int
-  """ legacy support when concerned with how many measurement tuple elements are used by user  """
-  _measCount::Int
-  """ legacy suport for measurement sample values of old functor residual functions """
-  _legacyMeas::P
-  """ legacy support for variable values old functor residual functions """
-  _legacyParams::X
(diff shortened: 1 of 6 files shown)


src/entities/FactorOperationalMemory.jl
```diff
@@ -109,52 +77,52 @@ DevNotes
...
+    _gradients::G
+    # type used for cache
+    dummyCache::CT
+
+    #Consolidation from FMD
+    fullvariables::Vector{DFGVariable}
+
+    #Consolidation from CPT
+    # the actual particle being solved at this moment
+    particleidx::Int
```
dehann, line 121: these Old CPT fields are specific to each particle (thread) in the future so can therefore not stay in CCW.  They should only be created in CalcFactor.

src/services/NumericalCalculations.jl
```diff
@@ -197,10 +197,10 @@ end
...
 
 Notes  
-- Unless passed in as separate arguments, this assumes already valid in `cpt_`:
-  - `cpt_.p`
-  - `cpt_.activehypo`
-  - `cpt_.factormetadata`
+- Unless passed in as separate arguments, this assumes already valid in `ccwl`:
+  - `ccwl.p`
+  - `ccwl.activehypo`
+  - `ccwl.factormetadata`
```
dehann, line 203: can remove this line

dehann (approved):
Good for removing CPT.  This PR does create the technical debt of using future thread specific memory in CCW.  So the old CPT fields will have to be removed from CCW again.

src/entities/FactorOperationalMemory.jl
Affie: We can discuss it at some stage. The mission here was to remove CPT.

src/services/NumericalCalculations.jl
Affie: done

Affie:
> Good for removing CPT. This PR does create the technical debt of using future thread specific memory in CCW. So the old CPT fields will have to be removed from CCW again.

Technical debt can be captured here #1616

src/entities/FactorOperationalMemory.jl
dehann: sounds good

dehann:
thanks Johan, you can merge this one too when you are ready.

PR #1617, opened 2022-08-29
Test Point2Point2Range as in RoME

Affie:
Basically the same test that fails in https://github.com/JuliaRobotics/RoME.jl/issues/618

### DIFF
--- a/src/services/EvalFactor.jl
+++ b/src/services/EvalFactor.jl
@@ -182,13 +182,13 @@ function computeAcrossHypothesis!(ccwl::Union{<:CommonConvWrapper{F},
       for i in 1:Threads.nthreads()  ccwl.cpt[i].activehypo = vars; end
       
       addEntr = view(ccwl.params[sfidx], allelements[count])
-      # dynamic estimate with user requested speadNH of how much noise to inject (inflation or nullhypo)
-      spreadDist = calcVariableDistanceExpectedFractional(ccwl, sfidx, certainidx, kappa=ccwl.inflation)
       
       # do proposal inflation step, see #1051
       # consider duplicate convolution approximations for inflation off-zero
       # ultimately set by dfg.params.inflateCycles
       for iflc in 1:inflateCycles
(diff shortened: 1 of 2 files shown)


src/services/EvalFactor.jl
```diff
@@ -182,13 +182,13 @@ function computeAcrossHypothesis!(ccwl::Union{<:CommonConvWrapper{F},
...
       addEntr = view(ccwl.params[sfidx], allelements[count])
-      # dynamic estimate with user requested speadNH of how much noise to inject (inflation or nullhypo)
-      spreadDist = calcVariableDistanceExpectedFractional(ccwl, sfidx, certainidx, kappa=ccwl.inflation)
       
       # do proposal inflation step, see #1051
       # consider duplicate convolution approximations for inflation off-zero
       # ultimately set by dfg.params.inflateCycles
       for iflc in 1:inflateCycles
+        # dynamic estimate with user requested speadNH of how much noise to inject (inflation or nullhypo)
+        spreadDist = calcVariableDistanceExpectedFractional(ccwl, sfidx, certainidx, kappa=ccwl.inflation)
```
dehann, line 191: I had it consistent over inflate cycles to ensure inflation stays strong rather than anneal down, are you trying to reduce the "noise" on range only results, or increase the inflation strength?  Perhaps just give a bit more detail on why this change?

dehann (changes requested):

src/services/EvalFactor.jl
Affie: increase the inflation strength: I do not know what changed, but it doesn't get there with a reasonable amount of `inflateCycles` 
But at least we know the problem has something to do with inflation.

Affie: If the variance is small to start with `spreadDist` is small and inflation takes a long time to "walk around the circle"

dehann: thanks Johan you're right

dehann (approved):
Go for merge, thanks

PR #1622, opened 2022-08-31
impr type stability on CalcFactor

dehann:


Affie:
I left it like this on purpose as it doesn't seem to make a difference as types are still dynamically dispatched and this just increases the compile time.

Affie:
You can go ahead and merge if you are seeing a performance improvement.

dehann:
given JL 1.8 and SnoopPrecompile, compile times are no longer a consideration.  If you noticed longer compile times then there should be performance benefits given stronger types.  All we should do is make sure the main body of dispatches are captured during precompile.  going to merge this thanks.

Affie:
Perhaps just post your benchmark code here for this improvement, or perhaps it's better if we make a separate benchmark issue.
I wan't to start collecting benchmark tests to include in CI at a later stage.

PR #1601, opened 2022-08-18
Fix julia 1.8 bug on tree deepcopy

Affie:
close https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1530

This patch might only be temporary if it is indeed a julia bug.

Issue #1530, opened 2022-06-02
MethodError: no method matching deepcopy_internal on julia 1.8 rc1


Affie:
In Test solveCliqueUp! and solveCliqDown! at ../test/testCliqSolveDbgUtils.jl:6

```julia
ERROR: MethodError: no method matching deepcopy_internal(::ReentrantLock)
Closest candidates are:
  deepcopy_internal(::Base.AbstractLock, ::IdDict) at deepcopy.jl:130
  deepcopy_internal(::Any, ::IdDict) at deepcopy.jl:53
  deepcopy_internal(::BigInt, ::IdDict) at gmp.jl:794
  ...
Stacktrace:
  [1] deepcopy_internal(x::Base.GenericCondition{ReentrantLock}, stackdict::IdDict{Any, Any})
    @ Base ./deepcopy.jl:143
  [2] deepcopy_internal(x::Any, stackdict::IdDict{Any, Any})
    @ Base ./deepcopy.jl:65
  [3] deepcopy_internal(x::Dict{Symbol, Any}, stackdict::IdDict{Any, Any}) (repeats 2 times)
    @ Base ./deepcopy.jl:125
  [4] deepcopy_internal(x::Any, stackdict::IdDict{Any, Any}) (repeats 3 times)
    @ Base ./deepcopy.jl:65
  [5] deepcopy
    @ ./deepcopy.jl:26 [inlined]
  [6] (::StateMachine{CliqStateMachineContainer})(userdata::CliqStateMachineContainer{BayesTreeNodeData, GraphsDFG{SolverParams, DFGVariable, DFGFactor}, GraphsDFG{SolverParams, DFGVariable, DFGFactor}, MetaBayesTree}, timeout::Nothing; pollinterval::Float64, breakafter::Function, verbose::Bool, verbosefid::Base.TTY, verboseXtra::IncrementalInference.CliqStatus, iterlimit::Int64, injectDelayBefore::Nothing, recordhistory::Bool, housekeeping_cb::IncrementalInference.var"#479#481"{IncrementalInference.TreeClique, CliqStateMachineContainer{BayesTreeNodeData, GraphsDFG{SolverParams, DFGVariable, DFGFactor}, GraphsDFG{SolverParams, DFGVariable, DFGFactor}, MetaBayesTree}})
    @ FunctionalStateMachine ~/.julia/packages/FunctionalStateMachine/2JZFG/src/StateMachine.jl:76
[...]

PR #1629, opened 2022-09-11
restore jl deepcopy_interal #1530 and #1601

dehann:


dehann:
tested local and works.  Actually need both the recent small PRs to continue working so going to merge prematurely.  If we find an issue with this deprecation we can just fix in a new PR before tagging.

PR #1643, opened 2022-12-11
Cleanup calcCovarianceBasic

Affie:


Affie:
I think test failure is not related to this PR and there might have been a regression somewhere.

PR #1650, opened 2022-12-16
Fix parametric test regression 

### LINKED ISSUE #1648: New test failure in parametric mixture
`test parametric mixture: Test Failed at /home/runner/work/IncrementalInference.jl/IncrementalInference.jl/test/testMixtureParametric.jl:54`

Affie:
close #1648

### DIFF
--- a/src/ParametricUtils.jl
+++ b/src/ParametricUtils.jl
@@ -509,14 +509,8 @@ function solveGraphParametric(
   computeCovariance::Bool = true,
   solveKey::Symbol = :parametric,
   autodiff = :forward,
-  algorithm = Optim.NewtonTrustRegion, # Optim.BFGS,
-  algorithmkwargs = (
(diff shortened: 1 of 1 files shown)


src/ParametricUtils.jl
```diff
@@ -509,14 +509,8 @@ function solveGraphParametric(
   computeCovariance::Bool = true,
   solveKey::Symbol = :parametric,
   autodiff = :forward,
-  algorithm = Optim.NewtonTrustRegion, # Optim.BFGS,
```
Affie, line 512: @dehann, just use the `algorithm` and `algorithmkwargs` keyword arguments if you want to use a different algorithm.

dehann: alright, thanks.  I was testing stuff here.  Didn't realize you had explored the different algorithms to that extent yet.  Thanks for the fix

Issue #1661, opened 2022-12-28
Confirm ccw.measurement for prior or relative (point or tangent)


dehann:
Usage of ccw.measurement is scattered and predates the decision to have getSample for priors or relative factors to return either points or tangents, respectively.

TAC

- Go through IIF code base and make sure all uses of ccw.measurements respect either point or tangent usage.  This will be a silent bug if not actively checked.

Issue #1492, opened 2022-02-13
Standarize partials definition and behavior


dehann:
We should do better than define partials as coordinate dimension positions -- i.e. get away away from say a Yaw factor partial is `(3,)` (as in coordinates).  Rather use a projection function or better.

dehann:
Going for partial factors' `getManifold` to only return the partial manifold definition; as well as `getSample` to return tangents or points on that partial factor.  The rest of the IIF logic should then cater for which variables those factors are connected to and update the numerics accordingly.  Currently that is done via factors having `.partials::Vector{Int}` which strictly specifies the partial dimensions.  This is not perfect and has a bunch of limitations in the near term, but good enough for transition to Manifolds proper so that partial manifolds can be broadly advanced as necessary.

PR #1662, opened 2022-12-28
add CalcFactor.manifold, all factors getManifold

dehann:
Largely replaces:
- #1415 

Also relates to:
- #1492

Helped discover a potential series of bugs:
- #1661


dehann:
This PR is showing bugs downstream with RoME tests on PartialPose3ZRP.  Will first clear that before merging here.

Issue #1044, opened 2020-12-09
Document how to use CF data in factor evaluation


dehann:
taking over from #783 

xref #467

dehann:
```julia
function (cf::CalcFactor<:LinearRelative})(res, z, x1, x2)
  # cf.metadata
  #  ...
end
```

dehann:
Minimum docs here: https://juliarobotics.org/Caesar.jl/latest/examples/basic_definingfactors/#Metadata

Affie:
FMD is deprecated. So we should document replacements is calcfactor

dehann:
I've added quite a bit of documentation to the types CCW and CalcFactor which should already help a lot.  Still worth adding a bit to Caesar docs on own factors just for quicker learning curve.

PR #1666, opened 2022-12-29
impr testing, better ccw field order towards multihypo checks

dehann:
get all the ccw fields into a logical sequence.  Get all multihypo related fields together.  Get all config parameters together.  Get all topologic fields together, etc.

Helps with:
- #1044


dehann:
Found a massive bug, somehow the `evalPotentialSpecific {<:AbstractFactor}` function got duplicated and updates to first definition got masked by second definition.  Will fix in next PR

PR #1678, opened 2023-01-27
Merging my fix branch into your v0.19 feature branch

GearsAD:
The create/update calls need to return the full variable/factor/solverdata/ppe etc. so that they have all the underlying information for subsequent calls. There may be more that we need to update.


dehann:
thanks, I'll will merge and resolve with more fixes in:
- #1677

PR #1690, opened 2023-03-04
tests not broken w dfg 20

dehann:


Affie (approved):
It won't pass with the current CI setup until DFG is tagged.

Project.toml
```diff
@@ -52,7 +52,7 @@ ApproxManifoldProducts = "0.6.3"
 BSON = "0.2, 0.3"
 Combinatorics = "1.0"
 DataStructures = "0.16, 0.17, 0.18"
-DistributedFactorGraphs = "0.19.3, 0.20"
+DistributedFactorGraphs = "0.20"
```
Affie, line 55:
```suggestion
DistributedFactorGraphs = "0.19.3 0.20"
```
Lets just leave 19 until 20 is tagged so tests can pass

PR #1691, opened 2023-03-08
NoSolverParams does not have attemptGradients

GearsAD:
To confirm: If the DFG has NoSolverParams, this should pass `true`?


Affie (approved):
Thanks for fixing it, I forgot to test upstream on this fix.

PR #1713, opened 2023-05-11
Change points to StaticArrays

Affie:
HeatmapSampler test fails repeatedly on github ci, but passes locally. So skipping in this pr with a new issue to fix the test
- see issue #1714


Affie:
@dehann, I'm merging this as is to try and keep branches from diverging too much. 
RoME and IIF tests passes locally, but had to skip one test for CI.

PR #1720, opened 2023-05-28
restore three reg tests from #1716

dehann:
fix half of rollover work from:
- #1716 

Still an issue remaining, see comment below


dehann:
restore different point count N on solve graph, also fix tests:
https://github.com/JuliaRobotics/IncrementalInference.jl/blob/8610923e4f7f6be28d99583c4e1600f6b33e56ee/test/testVariousNSolveSize.jl#L25

PR #1776, opened 2023-09-19
Fix getPointType of Position{N}

Affie:

Issue #1546, opened 2022-06-29
Improve Parametric Batch Solve Performance


Affie:
This issue is currently just a placeholder for performance-related issues around parametric batch solve.
```julia
using IncrementalInference #v0.29
using RoME #v0.19

##

fg = initfg()

addVariable!(fg, :x0, ContinuousScalar)
addVariable!(fg, :x1, ContinuousScalar)
addVariable!(fg, :x2, ContinuousScalar)
addVariable!(fg, :x3, ContinuousScalar)
addVariable!(fg, :x4, ContinuousScalar)
addVariable!(fg, :l1, ContinuousScalar)

addFactor!(fg, [:x0], Prior(Normal(0.0,0.1)))
addFactor!(fg, [:x0,:x1], LinearRelative(Normal(1.0,0.1)))
addFactor!(fg, [:x1,:x2], LinearRelative(Normal(1.0,0.1)))
addFactor!(fg, [:x2,:x3], LinearRelative(Normal(1.0,0.1)))
addFactor!(fg, [:x3,:x4], LinearRelative(Normal(1.0,0.1)))
addFactor!(fg, [:x0,:l1], LinearRelative(Normal(1.0,0.1)))
addFactor!(fg, [:x4,:l1], LinearRelative(Normal(-3.0,0.1)))

initAll!(fg)

@time r = IIF.solveGraphParametric!(fg)
# 0.033984 seconds (151.65 k allocations: 6.195 MiB)
# Seconds run:   0  (vs limit 100)
# Iterations:    6
# f(x) calls:    13
# ∇f(x) calls:   13

##

fg = initfg()

addVariable!(fg, :x0, Point3)
addVariable!(fg, :x1, Point3)
addVariable!(fg, :x2, Point3)
addVariable!(fg, :x3, Point3)
addVariable!(fg, :x4, Point3)
addVariable!(fg, :l1, Point3)

addFactor!(fg, [:x0], PriorPoint3(MvNormal([0.,0,0], [0.1,0.1,0.1])))
addFactor!(fg, [:x0,:x1], Point3Point3(MvNormal([1.,0,0], [0.1,0.1,0.1])))
addFactor!(fg, [:x1,:x2], Point3Point3(MvNormal([1.,0,0], [0.1,0.1,0.1])))
addFactor!(fg, [:x2,:x3], Point3Point3(MvNormal([1.,0,0], [0.1,0.1,0.1])))
[...]

PR #1778, opened 2023-10-05
Refactor Parametric solve for better performance and better use of Manopt.jl

Affie:
Progress tracked in https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1546
Needs: https://github.com/JuliaRobotics/IncrementalInference.jl/pull/1776


test/testSphereMani.jl
```diff
@@ -14,6 +14,7 @@ import Manifolds: identity_element
 #FIXME REMOVE! this is type piracy and not a good idea, for testing only!!!
 Manifolds.identity_element(::Sphere{2, ℝ}) = SVector(1.0, 0.0, 0.0)
 Manifolds.identity_element(::Sphere{2, ℝ}, p::AbstractVector) = SVector(1.0, 0.0, 0.0) # Float64[1,0,0]
+DFG.getPointIdentity(::Sphere{2, ℝ}) = SVector(1.0, 0.0, 0.0)
```
dehann, line 17: perhaps just add a comment that this is an abuse of notation  (Sphere not being a group).  

I have been thinking about in future perhaps we should change this call to `getPointDefault` or `getPointFallback` and avoid use of identity on this context.  The reason for waiting on such a signature chance for me is first completing the solver upgrades to full Riemannian first and then come back.

src/parametric/services/ParametricUtils.jl
```diff
@@ -294,34 +310,32 @@ function GraphSolveBuffers(@nospecialize(M), ::Type{T}) where {T}
...
-struct GraphSolveContainer
+struct GraphSolveContainer{CFT}
   M::AbstractManifold # ProductManifold or ProductGroup
   buffers::OrderedDict{DataType, GraphSolveBuffers}
   varTypes::Vector{DataType}
   varTypesIds::OrderedDict{DataType, Vector{Symbol}}
-  cfdict::OrderedDict{Symbol, CalcFactorMahalanobis}
   varOrderDict::OrderedDict{Symbol, Tuple{Int, Vararg{Int}}}
-  # cfarr::AbstractVector # TODO maybe <: AbstractVector(CalcFactorMahalanobis)
+  cfv::ArrayPartition{CalcFactorMahalanobis, CFT}
```
dehann, line 319:
```suggestion
  cfv::ArrayPartition{<:CalcFactorMahalanobis, CFT}
```

Affie: The ArrayPartition should be able to store a mixture of any CalcFactor type, so the `<:` will not work here.

Affie: I'm not actually sure if the first parameter is used at all. 
I'm thinking of moving the optim.jl solve to the attic in any case.

Issue #1420, opened 2021-10-15
Refactor Mixture as direct Density


dehann:
Another question here is whether we should just update Mixture to be a regular `<:SamplableBelief` density rather than quasi factor-density structure it is now.  i.e. change everything so that `PriorPose2(Mixture((.., ...), (0.5;0.5)))`.  Should never have gone down the road of `Mixture(PriorPose2, (..,...),(0.5,0.5))` as it is now.

_Originally posted by @dehann in https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1407#issuecomment-944361944_

dehann:
From later duplicate:

### Project Epic board

https://github.com/orgs/JuliaRobotics/projects/5/views/1

### Discussed in https://github.com/JuliaRobotics/Caesar.jl/discussions/808

Continue chatting on Discussion page, this Issue is just boiler plate link to Project Epics.  
( we can unlock this conversation if necessary ).

dehann:
Also see discussion at:
- JuliaRobotics/DistributedFactorGraphs.jl#1133

Affie:
We can also consider using Distributions.jl by expanding MixtureModel if we implement `Sampleable`.
https://github.com/JuliaRobotics/DistributedFactorGraphs.jl/issues/1146#issuecomment-3033245065

Also:
https://github.com/JuliaStats/Distributions.jl/blob/abb151c42eb67b4994e21da15e24c65f1cc3dd9f/src/mixtures/mixturemodel.jl#L24

PR #1779, opened 2023-10-11
Fix mixture sampling and deconv for static arrays

Affie:


src/Factors/Mixture.jl
```diff
@@ -133,10 +133,19 @@ function sampleFactor(cf::CalcFactor{<:Mixture}, N::Int = 1)
   #out memory should be right size first
   length(cf.factor.labels) != N ? resize!(cf.factor.labels, N) : nothing
   cf.factor.labels .= rand(cf.factor.diversity, N)
+  M = cf.manifold
+  
+  # mixture needs to be refactored so let's make it worse :-)
```
dehann, line 138: we're going to ditch `.mechanics` with Mixture refactor:
- #1420

src/entities/AliasScalarSampling.jl
```diff
@@ -46,6 +46,14 @@ struct AliasingScalarSampler
   end
 end
 
+function sampleTangent(
+  M::AbstractDecoratorManifold,
```
dehann, line 50:
```suggestion
  M::AbstractDecoratorManifold, # stand-in type to restrict to just group manifolds
```

Issue #1781, opened 2023-10-11
dev issue RoME tests with manopt


dehann:
since mergining:
- #1778 

On `using RoME`

```julia
ERROR: [ Info: monitorCSMs: all tasks done
LoadError: type CalcFactorResidual has no field manifold
Stacktrace:
  [1] getproperty
    @ Base ./Base.jl:37 [inlined]
  [2] (::IncrementalInference.CalcFactorResidual{RoME.PriorPose2{Distributions.FullNormal}, Nothing, 3, 9, SubArray{Float64, 1, RecursiveArrayTools.ArrayPartition{Float64, Tuple{Vector{RecursiveArrayTools.ArrayPartition{Float64, Tuple{StaticArraysCore.SVector{2, Float64}, StaticArraysCore.SMatrix{2, 2, Float64, 4}}}}, Vector{StaticArraysCore.SVector{2, Float64}}}}, Tuple{Vector{Int64}}, false}, RecursiveArrayTools.ArrayPartition{Float64, Tuple{StaticArraysCore.SVector{2, Float64}, StaticArraysCore.SMatrix{2, 2, Float64, 4}}}, 1})(m::RecursiveArrayTools.ArrayPartition{Float64, Tuple{StaticArraysCore.SVector{2, Float64}, StaticArraysCore.SMatrix{2, 2, Float64, 4}}}, p::RecursiveArrayTools.ArrayPartition{Float64, Tuple{StaticArraysCore.SVector{2, Float64}, StaticArraysCore.SMatrix{2, 2, Float64, 4}}})
    @ RoME ~/.julia/dev/RoME/src/factors/PriorPose2.jl:41
[...]

PR #1782, opened 2023-10-12
workaround hack for #1781

dehann:
see
- #1781


dehann:
Hi, @Affie , could you take over this fix -- I just pushed up the workaround I put in place but need your input on which way you wanted to go with `CalcFactor.manifold` question

PR #1783, opened 2023-10-12
cf.manifold for CalcFactor

dehann:


Affie (approved):
I'm not sure about the performance of overloading getproperty, but this will be temporary.

PR #1784, opened 2023-10-12
Swop parametric optim for manopt and better convolution performance 

Affie:


src/parametric/services/ParametricManopt.jl
```diff
@@ -297,7 +297,12 @@ function covarianceFiniteDiff(M, jacF!::JacF_RLM!, p0)
     H = FiniteDiff.finite_difference_hessian(costf, X0)
 
     # inv(H)
-    Σ = Matrix(H) \ Matrix{eltype(H)}(I, size(H)...)
+    if isapprox(det(H), 0, atol=eps())
```
dehann, line 300: FYI `det` is a heavy computation, so will be slow for large matrices

PR #1786, opened 2023-10-13
Use CalcFactorNormSq as functor

### LINKED ISSUE #1785: approxDeconv result wraps around on circular manifolds
In #1778 (or upstream) something (possibly related to upgrading to StaticArrays) changed that caused the result to wrap around. It is still correct though.

Affie:
~~This only focused on hex (pose2 and point2) a lot of other things are untested and likely broken.~~
Tests are looking ok.

This also fixes a deconv issue where the variable manifold was used instead of the factor manifold. and close #1785

### DIFF
--- a/src/entities/CalcFactor.jl
+++ b/src/entities/CalcFactor.jl
@@ -35,7 +35,8 @@ struct CalcFactorNormSq{
   X, 
   C, 
   VT <: Tuple, 
-  M <: AbstractManifold
+  M <: AbstractManifold,
+  MEAS
 } <: CalcFactor{FT}
   """ the interface compliant user object functor containing the data and logic """
   factor::FT
@@ -43,7 +44,7 @@ struct CalcFactorNormSq{
   _sampleIdx::Int
   """ legacy support for variable values old functor residual functions.
       TBD, this is still being used by DERelative factors. """
-  _legacyParams::X
+  _legacyParams::X #TODO rename to varValsHypo for consistent naming? and not not legacy any more
   """ allow threading for either sampling or residual calculations (workaround for thread yield issue) """
   _allowThreads::Bool
   """ user cache of arbitrary type, overload the [`preambleCache`](@ref) function. NOT YET THREADSAFE """
@@ -56,6 +57,7 @@ struct CalcFactorNormSq{
   # which index is being solved for?
   solvefor::Int
   manifold::M
+  measurement::MEAS
 end
 
 #TODO deprecate after CalcFactor is updated to CalcFactorNormSq
--- a/src/services/CalcFactor.jl
+++ b/src/services/CalcFactor.jl
@@ -43,7 +43,8 @@ function CalcFactorNormSq(
     cache,
     tuple(fullvariables...),
     solvefor,
-    manifold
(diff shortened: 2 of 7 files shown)


src/services/CalcFactor.jl
```diff
@@ -43,7 +43,8 @@ function CalcFactorNormSq(
     cache,
     tuple(fullvariables...),
     solvefor,
-    manifold
+    manifold,
+    ccwl.measurement,
```
dehann, line 47: This is all samples?

src/entities/CalcFactor.jl
```diff
@@ -35,15 +35,16 @@ struct CalcFactorNormSq{
...
+  MEAS
 } <: CalcFactor{FT}
   """ the interface compliant user object functor containing the data and logic """
   factor::FT
   """ what is the sample (particle) id for which the residual is being calculated """
   _sampleIdx::Int
   """ legacy support for variable values old functor residual functions.
       TBD, this is still being used by DERelative factors. """
-  _legacyParams::X
+  _legacyParams::X #TODO rename to varValsHypo for consistent naming? and not not legacy any more
```
dehann, line 47: One CalcFactor per sample?  History idea was CCW per solve and heap CalcFactor per sample.  Then we can do multithreading for each sample.  Also for more complicated factors that require own cache memory location.  Don't want to index into cf.somemem[sampleIdx]

src/services/NumericalCalculations.jl
```diff
@@ -452,6 +370,43 @@ end
...
+  # set the target hypo on the correct sample to free variable x, was target object
+  varValsHypo[cf.solvefor][sampleIdx] = x
+
+  res = cf(cf.measurement[sampleIdx], map(vvh -> _getindex_anyn(vvh, sampleIdx), varValsHypo)...)
+  return sum(x->x^2, res)
+end
+
+function _buildHypoCalcFactor(ccwl::CommonConvWrapper, smpid::Integer)
+  # build a view to the decision variable memory
+  varValsHypo = ccwl.varValsAll[][ccwl.hyporecipe.activehypo]
```
dehann, line 398: Looks good, I just want talk through the one CalcFactor for each sample solve to be sure we're on the right track

dehann (commented):
Mostly looks good but want to talk through idea of each sample gets own CalcFactor with own .cache.  This also so that the ensemble solve of points can easily be made multithreaded in future

Affie:
A new calcfactor still gets created for each sample, but it currently shares some elements such as measurement.

src/services/CalcFactor.jl
Affie: Yes, currently

dehann:
> A new calcfactor still gets created for each sample, but it currently shares some elements

Okay, great -- this looks good to me then thanks!

dehann (approved):

Affie:
Will do the same for deconv in a separate PR

Affie:
NOTE `_slack` is not implemented for the new calc factor approach and will not work on ManifoldMinimize yet.
Should it or is autodiff the replacement?

PR #1793, opened 2023-10-20
rm getprop on cf.manifold

dehann:
we're probably not going for a dep warning as this was never released


Affie (approved):

Affie:
This looks good to me, we can merge this and then make sure it's not used downstream in RoME and Caesar.

PR #1798, opened 2023-10-25
min compat Manopt v0.4.40 and ManifoldsBase v0.15 fixes

dehann:


Project.toml
```diff
@@ -80,8 +80,8 @@ JSON3 = "1"
 KernelDensityEstimate = "0.5.6"
 ManifoldDiff = "0.3"
 Manifolds = "0.8.15, 0.9"
-ManifoldsBase = "0.13.12, 0.14, 0.15"
-Manopt = "0.4.27"
+ManifoldsBase = "0.14, 0.15"
+Manopt = "0.4.40"
```
Affie, line 84: It looks like we should drop ManifoldsBase v0.8 completely

dehann: sounds good

Issue #1807, opened 2023-12-09
Proposal for robust parametric kernels with null hypo


dehann:
```julia
using GLMakie
using LinearAlgebra

##

L(r, δ_) = δ_^2 * (sqrt(1 + (r/δ_)^2) - 1)

# try: δ = im*r / σ^2
δ(r, σ; nh=0) = (1-nh)^4*im*r / σ^2 + (nh/(nh+1e-5))/(σ^2)
L_(r, σ; nh=0) = L(r, δ(r,σ;nh))

XX = -30:0.1:30
nh = 0.2

f = Figure()
ax = Axis(f[1,1];
  title="""
  L(r, δ_) = δ_^2 * (sqrt(1 + (r/δ_)^2) - 1)   # PseudoHuber loss
  Posit: δ = im*r / σ^2                  # dehann01
  δ(r, σ; nh=0) = (1-nh)^4*im*r / σ^2 + (nh/(nh+1e-5))/(σ^2)
  L_(r, σ; nh=0) = L(r, δ(r,σ;nh))""",
  xgridcolor = :gray,
  ygridcolor = :gray,
  xgridwidth = 1,
  ygridwidth = 1,
  xminorgridcolor = :gray,
  yminorgridcolor = :gray,
  xminorgridvisible = true,
  yminorgridvisible = true,
)
lines!(ax,XX, 1 .+ (XX.^2), color=:cyan, label="σ=1,r^2")
lines!(ax,XX, norm.(L_.(XX, 1.0)), color=:blue, label="σ=1,nh=0")
lines!(ax,XX, norm.(L_.(XX, 1.0; nh)), color=:red, label="σ=1,nh=$nh")
lines!(ax,XX, norm.(L_.(XX, 2.0)), color=:green, label="σ=2,nh=0")
lines!(ax,XX, norm.(L_.(XX, 2.0; nh)), color=:magenta, label="σ=2,nh=$nh")
lines!(ax,XX, norm.(L_.(XX, 1.0; nh=1.0)), color=:orange, label="σ=1,nh=1.0")
lines!(ax,XX, norm.(L_.(XX, 2.0; nh=1.0)), color=:brown, label="σ=2,nh=1.0")

axislegend()

# pl = lines(XX, real.(YY) .- XX.^2)
# lines!(XX, imag.(δ.(XX, 1.0; nh)), color=:tomato)

f
```

[...]

PR #1808, opened 2023-12-09
prototype nullhypo parametric with robust kernel

dehann:
relates to:
- #1807


Affie:
Are you suggesting that we revive the optim.jl solver code? This is currently just kept until the tree solve and mixtures can be updated to manopt. As far as i know it’s not used anywhere. 
I had it that CalcFactorResidual would replace CalcFactorMahalanobus.

dehann:
Hi Affie, no lets go for Manopt.  Was just easier to start the coding cycle against coordinates here.  We can copy from this code for future work.

[sic] Malahanobis

dehann:
Note that we have not added this into main via another PR just yet -- still TODO.

PR #1825, opened 2024-05-06
Experiment with factor gradients in solve

Affie:
Draft PR not to lose possible value

### DIFF
--- a/IncrementalInference/src/parametric/services/ParametricManopt.jl
+++ b/IncrementalInference/src/parametric/services/ParametricManopt.jl
@@ -72,10 +72,15 @@ function CalcFactorResidualAP(
   return ArrayPartition{CalcFactorResidual, typeof(parts_tuple)}(parts_tuple)
 end
 
-function (cfm::CalcFactorResidual)(p)
+function (cfm::CalcFactorResidual)(p::Vector)
   meas = cfm.meas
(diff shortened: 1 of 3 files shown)


src/parametric/services/ParametricManopt.jl
```diff
@@ -62,12 +62,17 @@ function CalcFactorResidualAP(fg::GraphsDFG, factorLabels::Vector{Symbol}, varIn
...
-function (cfm::CalcFactorResidual)(p)
+function (cfm::CalcFactorResidual)(p::Vector)
   meas = cfm.meas
   points = map(idx->p[idx], cfm.varOrderIdxs)
   return cfm.sqrt_iΣ * cfm(meas, points...)
 end
 
+function  (cfm::CalcFactorResidual)(p::ArrayPartition)
+  points = map(idx->p.x[idx], cfm.varOrderIdxs)
+  return cfm.sqrt_iΣ * cfm(cfm.meas, points...)
```
dehann, line 73: Post AMP 41, we should look at consolidating this with the kernel evaluations done in AMP.  Specifically this Malahanobis distance part.  The reasoning is that we need to build belief trees on each clique and then be able to evaluation, which is exactly the same requirement as what this code here is doing.

PR #1851, opened 2024-12-01
update for ManifoldDiff v0.4 backend

dehann:


dehann (commented):
Also added the basic fix in remote issue:
- https://github.com/JuliaManifolds/ManifoldDiff.jl/issues/26#issuecomment-2509617717

PR #1859, opened 2025-04-29
use QR for linear subsolver in parametric RLM

Affie:


src/parametric/services/ParametricManopt.jl
```diff
@@ -354,6 +360,7 @@ function solve_RLM(
     jacobian_tangent_basis = DefaultOrthogonalBasis(),
     initial_residual_values,
     initial_jacobian_f,
+    linear_subsolver!,
```
Affie, line 363: @mateuszbaran, can you please help me with how to use a custom `linear_subsolver!` in `LevenbergMarquardt!`. I couldn't quite figure out where to pass it in. QR is more numerically stable for SLAM problems so would like to try it out. Also note, ldiv! has an issue with sparse arrays but is fixed in julia 1.12. 
"MethodError: no method matching ldiv!(::SparseArrays.CHOLMOD.Factor{Float64, Int64}, ::Vector{Float64})"

Affie: I might also be wrong here with using `LevenbergMarquardt!` directly?

mateuszbaran: Ah, I forgot to put the subsolver as a keyword argument in the `LevenbergMarquardt!` function. I will add it tomorrow.

mateuszbaran: Manopt 0.5.14 (just tagged) has the keyword argument. Let me know if you encounter any further issues.

dehann: Thanks!

PR #1872, opened 2025-06-20
Splitting out IIFTypes as a separate module.

Affie:
This pr should split out: 
- some variable types + serialization
- some factor types + serialization
- some packed distribution types + serialization

Serialization is pack/unpack only
IIFTypes should be ready to register as subproject (won't register yet)
DFG + IIFTypes [later RoMETypes] should be able to pack/unpack variables and factors without IIF.


Affie:
Because mixtures do not follow the standard pattern, we need to resolve https://github.com/JuliaRobotics/IncrementalInference.jl/issues/1420 before IIFTypes can be fully supported/implemented.

PR #1878, opened 2025-07-25
fast-forward keep solve changes (first use is extracting DERelative odes)

dehann:
Hi @mei-cheung , here is the stuff you need.  Will be merging now


dehann:
Hi @Affie , going to merge here.  I tested on a local twig branch for conflicts [w liegroups branch] -- there was one small conflict that is easy to fix.  The conflict was a simple delete "where T" portion of a change you did somewhere, and keeping the new function argument `keepSolution`.

Affie:
If I’m not mistaken. The test failure has something to do with prior mixtures that broke with one of the recent DFG updates. So would agree that it’s unrelated to this pr.

PR #1882, opened 2025-09-12
Refactor Mixture as direct Density and standardize sampling on FluxModelsDistributions to resemble <: Sampleable

### LINKED ISSUE #1420: Refactor Mixture as direct Density

Affie:
Needs: https://github.com/JuliaRobotics/ApproxManifoldProducts.jl/pull/320

Close #1420

Note: 
- Sampleable beliefs need to work the same as <: Sampleable
- Only somewhat breaking, but mostly kept the Mixture signature.

### DIFF
--- a/IncrementalInference/src/Factors/Mixture.jl
+++ b/IncrementalInference/src/Factors/Mixture.jl
@@ -1,6 +1,3 @@
-
-_defaultNamesMixtures(N::Int) = ((Symbol[Symbol("c$i") for i = 1:N])...,)
-
 """
 $(TYPEDEF)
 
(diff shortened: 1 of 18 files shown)


dehann:
for readers finding this PR, note that we know IIF.Mixture needs to be refactored and streamlined.  We are working along a path of minimal breaking changes and `effort x max` compatibility.

IncrementalInference/src/Factors/Mixture.jl
```diff
@@ -12,195 +9,77 @@ Notes
...
-msp = Mixture(Prior, 
-              [Normal(0,0.1), Uniform(-pi/1,pi/2)],
-              [0.5;0.5])
+msp = Prior(Mixture([Normal(0,0.1), Uniform(-pi/1,pi/2)], [0.5;0.5])
 
 addFactor!(fg, [:head], msp, tags=[:MAGNETOMETER;])
 
 # Or relative
-mlr = Mixture(LinearRelative, 
+mlr = LinearRelative(Mixture(
```
dehann, line 22: yes!

dehann (approved):
Yes, thanks @Affie , this goes a long way to refactoring Mixture towards regular density!  Code looks good to me and also AMP 320

PR #1885, opened 2025-10-09
Remove PPEs - use calcMeanMaxSuggested

Affie:
This removes PPEs that will be obsolete in DFG v0.29 (marked as deprecated in v0.28). The next IIF can then deal with the struct breaking changes introduced in v0.29


IncrementalInference/src/Deprecated.jl
```diff
@@ -155,12 +155,218 @@ function sampleTangent(x::ManifoldKernelDensity, p = mean(x))
...
+  @info "Adding simulated loop closure with perfect data association" :x5 genLabel
+  addFactor!(fg, [:x5; genLabel], newFactor)
+end
+
+# the point is that only the (0,20) values in newFactor are needed, all calculations are abstracted away.
+```
+
+See also: [`RoME.generateGraph_Honeycomb!`](@ref), [`accumulateFactorMeans`](@ref)
+"""
+function _checkVariableByReference(
```
dehann: Why are these moving to deprecated?  These are used by the graph generators in RoME.

IncrementalInference/src/parametric/services/ParametricUtils.jl
```diff
@@ -973,7 +965,7 @@ function createMvNormal(v::VariableCompute, key = :parametric)
     dims = vnd.dims
     return createMvNormal(vnd.val[1:dims, 1], vnd.bw[1:dims, 1:dims])
   else
-    @warn "Trying MvNormal Fit, replace with PPE fits in future"
+    @warn "Trying MvNormal Fit, replace with homotopy? fits in future"
```
dehann:
```suggestion
    @warn "Trying MvNormal Fit"
```

dehann (changes requested):
looks good except for `_checkVariablesByReference`?

IncrementalInference/src/Deprecated.jl
Affie: See my note above:

```
#FIXME The next functions use PPEs and should be updated or deprecated
# getPPESuggestedAll no external use
# findVariablesNear used in 1 rome example
# _checkVariableByReference used in ROME #TODO
```

But, you are right. I should maybe not move it to deprecated. I didn't want to go down a rabbit hole to fix it, but should maybe try so this can form part of the next release.

Affie: I moved it out of deprecated, but `calcMeanMaxSuggested` is based on what is stored in `State` and downstream code will have to be updated.

dehann (approved):
sure, TODO broken seem like a better path.  It's indicated by underscore as internal -- don't think it's exported.  I'm happy to merge as is.

PR #1906, opened 2026-03-17
Composite rotations.

samuelsonric:
This PR allows you to use any ordering method with contraints (rather than just `colamd`).

I cannot figure out how to test it: the library will not compile.


dehann (approved):
Hi @samuelsonric , thanks for the contribution -- code looks good and tests pass locally.  Going to merge via alternate PR #1926 which has fast-forwarded changes required for tests to pass.  These commits will be available in v0.37.0 soon.

PR #1915, opened 2026-04-28
Restore ForwardDiff support in parametric and use precision matrices

Affie:


Affie:
Dev tests now pass again.

PR #1924, opened 2026-05-19
Towards removing SolverParams from DFG

Affie:
WIP on splitting SolverOptions and DebugOptions:
```julia
@kwdef struct SolverOptions
  """ Quasi fixed length """
  qfl::Int = typemax(Int)
  """ true when adhering to qfl window size for solves """
  isfixedlag::Bool = false
  """ if true, then fixed lag will not update marginalized during down pass on tree """
  limitfixeddown::Bool = false
  """ use incremental tree updates, TODO consolidate with recycling """
  incremental::Bool = true
  """ Experimental, insert differential factors from upward joints """
  useMsgLikelihoods::Bool = false
  """ do tree upsolve """
  upsolve::Bool = true
  """ do tree downsolve """
  downsolve::Bool = true
  """ limit number of steps CSMs can take """
  limititers::Int = 500
  """ default number of particles """
  N::Int = 100
  """ default to graph-based initialization of variables """
  graphinit::Bool = true
  """ init variables on the tree """
  treeinit::Bool = false
  limittreeinit_iters::Int = 10
  """ entropy spread adjustment used for both null hypo cases. """
  spreadNH::Float64 = 3.0
  """ how much to disperse particles before convolution solves, #1051 """
  inflation::Float64 = 5.0
  """ minimum nullhypo for relative factors sibling to multihypo factors onto a specific variable. """
  nullSurplusAdd::Float64 = 0.3
  """ repeat convolutions for inflation to occur """
  inflateCycles::Int = 3
  """ number of Gibbs cycles to take per clique iteration variables """
  gibbsIters::Int = 3
  """ maximum incidence to a variable in an effort to enhance sparsity """
  maxincidence::Int = 500
  """ Development feature on whether new samples should be sampled at each Gibbs cycle convolution """
  alwaysFreshMeasurements::Bool = true
  """ should factor gradients be calculated or attempted (UNDER DEVELOPMENT, 21Q3) """
  attemptGradients::Bool = false
  # These are best passed directly as kwargs into solveGraph! or housed in an ephemeral ExecutionOptions struct
  """ do not block on CSM tasks """
  async::Bool = false #TODO DFG pattern looks like it will rather be solveGraphAsync!
  """ should Distributed.jl tree solve compute features be used """
  multiproc::Bool = 1 < nprocs()
end

@kwdef struct DebugOptions
    """ Experimental, enable additional tier debug features """
    dbg::Bool = false
    """ "/tmp/caesar/logs/$(now())" # unique temporary file storage location for a solve """
    logpath::String = joinpath(tempdir(),"caesar","logs","$(now(DFG.UTC))")
      """ draw tree during solve """
    drawtree::Bool = false
    """ show CSM iteration count on tree visualization """
    drawCSMIters::Bool = true
    showtree::Bool = false
    """ how fast should the tree vis file be redrawn """
    drawtreerate::Float64 = 0.5
end
```


IncrementalInference/src/parametric/services/ParametricManopt.jl
```diff
@@ -677,14 +744,28 @@ function autoinitParametric!(
...
     end
 
+    if neighbor_seed
+      has_prior = any(isPrior.(dfg, listNeighbors(dfg, initme)))
+      if !has_prior && !isempty(initfrom)
+        # seed from the first initialized neighbor of the same variable type
+        my_kind = getStateKind(xi)
+        same_kind = filter(vl -> getStateKind(getVariable(dfg, vl)) === my_kind, initfrom)
+        if !isempty(same_kind)
+          DFG.refMeans(vnd)[1] = DFG.refMeans(getState(dfg, same_kind[1], solveKey))[1]
```
dehann: reminder to change to HomotopyDensity interface instead

IncrementalInference/src/parametric/services/ParametricManopt.jl
```diff
@@ -677,14 +744,28 @@ function autoinitParametric!(
...
+        end
+        # else: keep current state as fallback
+      end
+      # if has_prior: keep current state — prior will drive the solve
+    end
+
     if perturb_point
       _M = getManifold(xi)
       p = DFG.refMeans(vnd)[1]
```
dehann: again HoDe instead

IncrementalInference/src/parametric/services/ParametricManopt.jl
```diff
@@ -699,7 +780,7 @@ function autoinitParametric!(
         )
       )
     end
-    M, vartypeslist, lm_r, Λ = solve_RLM_conditional(dfg, [initme], initfrom; solveKey, kwargs...)
+    M, vartypeslist, lm_r, Λ, _ = solve_RLM_conditional(dfg, [initme], initfrom; solveKey, linear_subsolver!,  kwargs...)
     
     val = lm_r[1]
     DFG.refMeans(vnd)[1] = val
```
dehann: again

IncrementalInference/src/services/FactorGraph.jl
```diff
@@ -809,97 +683,61 @@ end
...
-    Symbol(namestring),
-    varOrderLabels,
-    usrfnc,
-    hyper,
-    state,
-    solvercache;
-    tags = Set(union(tags, [:FACTOR])),
-    solvable,
-    timestamp = _zonedtime(timestamp),
+  newFactor = FactorDFG(
```
dehann, line 722: I'd like to at some point soonish like to review the DataLevel 3.5 idea of `const Factor = Union{<:FactorDFG, FactorCompute}`.

I'd also like to look again and noun `Compute` vs `Live`.  I currently favor `Live`, but that is because AMP v0.15 is working with this pattern, but want to make sure this idea scales before we drive it in everywhere.

IncrementalInference/src/services/FGOSUtils.jl
```diff
@@ -223,10 +224,7 @@ DevNotes
...
 - TODO standardize after AMP3D
 """
-function calcMeanMaxSuggested(
-  vari::VariableCompute,
-  solveKey::Symbol = :default
-)
+function calcMeanMaxSuggested(vari::VariableCompute, solveKey::Symbol = :default)
   varType = getStateKind(vari)
   P = getBelief(vari, solveKey)
   maniDef = convert(MB.AbstractManifold, varType)
```
dehann, line 230: why not just `getManifold(varType)`?

IncrementalInference/src/services/GraphInit.jl
```diff
@@ -487,13 +574,47 @@ function ensureSolvable!(
...
+
+Specifically:
+- Ensures every solvable variable has a `State` for `solveKey` with `N` particles (uninitialized).
+- Ensures every solvable factor has a built CCW (`solvercache`).
+
+No variable values are computed — use [`initAll!`](@ref) for that.
+
+See also: [`initAll!`](@ref), [`prepareFactorCache!`](@ref)
+"""
+function prepare!(
```
dehann, line 590: ah, this is forcing DataLevel4 rather than only-when-needed / JIT.

Also, any `prepare` should happen local to graph nodes to minimize concurrency/atomic races.

IncrementalInference/src/services/SolverAPI.jl
```diff
@@ -426,6 +435,7 @@ function solveTree!(
   algorithm != :parametric ? nothing : @error("Under development, do not use, see #539")
   !storeOld ? nothing : @error("parametric storeOld keyword not wired up yet.")
 
+  #TODO consider solveGraph! vs solveGraphAsync!
```
dehann, line 438: still not sure about `solveGraphAsync!` vs `@async solveGraph!`. 

Aside, CSM is the right place for asyncs, because that will help horizonal scalability of entire system down the road.  But again, why not just `@async solveClique!`?

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -7,7 +7,7 @@ using IncrementalInference
 
 @testset "Test consolidation of factors #467" begin
   fg = generateGraph_LineStep(20, poseEvery=1, landmarkEvery=4, posePriorsAt=collect(0:7), sightDistance=2, solverParams=SolverParams(algorithms=[:default, :parametric]))
-
+  IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 10: this new `prepare` step in user space needs to be folded into `solve` in the short term, and totally removed from UX as fast as possible.  I don't want this to become one of those things that never gets fixed because we are rushing now.

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -32,6 +32,7 @@ initVariable!(fg, :x0, Normal(0.1,1.1), :parametric)
 addVariable!(fg, :x1, ContinuousScalar)
 addFactor!(fg, [:x0,:x1], LinearRelative(Normal(1.0, 1.2)))
 
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 35: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -44,7 +45,7 @@ addVariable!(fg, :x2, ContinuousScalar)
 addFactor!(fg, [:x0,:x2], LinearRelative(Normal(2.0, 0.5)))
 addFactor!(fg, [:x1,:x2], LinearRelative(Normal(1.1, 0.5)))
 
-
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 48: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -62,7 +63,7 @@ end
 
 ##
 fg = generateGraph_LineStep(7, poseEvery=1, landmarkEvery=0, posePriorsAt=collect(0:7), sightDistance=2, solverParams=SolverParams(algorithms=[:default, :parametric]))
-
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 66: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -88,7 +89,7 @@ v0 = getVariable(fg,:x1)
 @test length(DFG.refMeans(v0.states[:parametric])[1]) === 1
 @test isapprox(DFG.refMeans(v0.states[:parametric])[1][1], 1.0, atol = 1e-4)
 
-
+initAll!(fg)
```
dehann, line 92: this is older but similar problem, it should not be in the UX

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -108,6 +109,7 @@ fg = generateGraph_LineStep(10, vardims=2, poseEvery=1, landmarkEvery=3, posePri
 # foreach(fct->println(fct.label, ": ", getObservation(fct).Z), getFactors(fg))
 
 # @profiler d,st = IIF.solveGraphParametric(fg)
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 112: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -175,7 +177,7 @@ addFactor!(fg, [:x1; :x2], LinearRelative(Normal(0.0, 1e-1)), graphinit=graphini
 
 
 foreach(fct->println(fct.label, ": ", getObservation(fct).Z), getFactors(fg))
-
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 180: again

IncrementalInference/test/testBasicParametric.jl
dehann: especially because it is solver specific, that makes it even worse

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -219,6 +216,7 @@ deleteFactor!(fg, :x5x6f1)
 # foreach(fct->println(fct.label, ": ", getObservation(fct).Z), getFactors(fg))
 
 # @profiler d,st = IIF.solveGraphParametric(fg)
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 219: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -261,6 +256,7 @@ end
 
 fg = generateGraph_LineStep(7, poseEvery=1, landmarkEvery=0, posePriorsAt=collect(0:7), sightDistance=2, solverParams=SolverParams(graphinit=false), graphinit=false)
 
+IIF.prepare!(fg, IIF.NLLSSolver(), :parametric)
```
dehann, line 259: again

IncrementalInference/test/testBasicParametric.jl
```diff
@@ -261,6 +256,7 @@ end
...
 @test (l->!isInitialized(fg, l, :parametric)).(ls(fg)) |> all
 
 initAll!(fg, :parametric)
```
dehann, line 262: again

IncrementalInference/test/TestCSMMultihypo.jl
```diff
@@ -57,7 +57,7 @@ addFactor!(fg, [:x2; :l2; :l2_0], LinearRelative(Normal(-40., meas_noise)), mult
 
 # drawGraph(fg)
 
-# initAll!(fg)
+initAll!(fg)
```
dehann, line 60: again

IncrementalInference/test/testMixtureLinearConditional.jl
```diff
@@ -89,6 +89,8 @@ f0 = addFactor!(fg, [:x0;], mp)
 mr = Mixture(LinearRelative, (fancy=manikde!(ContinuousEuclid(1), [randn(1) for _ in 1:75]), naive=Normal(0,10)), [0.4;0.6])
 f1 = addFactor!(fg, [:x0;:x1], mr)
 
+initAll!(fg)
```
dehann, line 92: again

IncrementalInference/test/testmultihypothesisapi.jl
```diff
@@ -70,6 +70,7 @@ v4 = addVariable!(fg, :x4, ContinuousScalar, N=N)
 ppMH = DevelopLikelihood(Normal(90.0,1.0))
 f3 = addFactor!(fg, [:x2;:x3;:x4], ppMH, multihypo=[1.0;0.5;0.5])
 
+initAll!(fg)
```
dehann, line 73: again

IncrementalInference/test/testmultihypothesisapi.jl
```diff
@@ -212,7 +213,7 @@ v5 = addVariable!(fg, :x5, ContinuousScalar, N=N)
 ppMH = DevelopLikelihood(Normal(90.0,1.0))
 f3 = addFactor!(fg, [:x2;:x3;:x4;:x5], ppMH, multihypo=[1.0,0.333,0.333,0.334])
 
-
+initAll!(fg)
```
dehann, line 216: again

IncrementalInference/test/testSaveLoadDFG.jl
```diff
@@ -10,7 +10,7 @@ using Test
 fg = generateGraph_Kaess()
 addVariable!(fg, :x4, ContinuousScalar)
 addFactor!(fg, [:x2;:x3;:x4], LinearRelative(Normal()), multihypo=[1.0;0.6;0.4])
-
+initAll!(fg)
```
dehann, line 13: again

IncrementalInference/test/testSolveKey.jl
```diff
@@ -36,8 +35,7 @@ initVariable!(fg, :a, pts, :testSolveKey)
 
 ##
 
-IIF.setDefaultNodeData!(getVariable(fg, :b), 0, 100; solveKey=:testSolveKey, 
-                        initialized=false, varType=ContinuousScalar())
+IIF.prepareState!(getVariable(fg, :b), IIF.NPBPSolver(), :testSolveKey; num_kernels=100)
```
dehann, line 38: yeah, really dont like this

IncrementalInference/test/testSpecialEuclidean2Mani.jl
```diff
@@ -579,7 +580,6 @@ end
 ##
 
 fg = initfg()
```
dehann, line 582: again even in legacy code

IncrementalInference/test/testUseMsgLikelihoods.jl
```diff
@@ -11,6 +11,7 @@ using IncrementalInference
 ##
 
 fg = generateGraph_CaesarRing1D()
+IIF.prepare!(fg, IIF.NPBPSolver(), :default)
```
dehann, line 14: again, nope

dehann (changes requested):
I'm pretty miffed about having to merge `prepare!` like it is in this PR. I want to make sure my STRONG protest against `prepare!` in UX, especially the algorithm specific prep, is captured.  This is definitely not the future state or design pattern.  The only reason here is a consolidation step, but know that `prepare!` (and legacy `initAll!`) must be removed -- like in yesterday.

I'm probably going to change to `const prepare_hack! = _prepare!` verbs to help force the issue.

Maybe I can motivate better:

The basic value prop of open-source data fusion library goes something like this:
- Good, transparent software that avoids everybody reimplementing the same basic needs, rather collaborate on common parts.
- State of the art features like solve recycling for large systems, or better real-time performance, hybrid(non)parametric solutions for overcoming contradictions.
- organic growth path to solicit help from commercial provider for either DX, compute/maintenance service, or offloading liability for a fee.
- UX/DX is what makes it "good".  UX is, i have observations and just want to throw them together and the AI figures it out:
```julia
fg = Local/RemoteFG()

# some cyclic block of code to ingest observations on an ongoing basis
for ob in observations
  addFactor!(fg, :autolabel, ob)
end
# note even addVariable! is friction here, since observations already give you the necessary structure 
#  ack, we punted the phantom/autovariable issue beyond DFG v1.0.

solveGraph!(fg; recycle=_uhm_)  # i.e. recycle previous compute in smart way

# extract results as part of value gained
ls(fg) .|> (l -> s=getBelief(fg,l); (l, mean(s), cov(s)))

# im guessing this type of user still has to learn about what `stateLabels` are for, so avoiding that above on purpose
```
- "Bad" UX implies the above example only works if you call a bunch of weird other functions (only at the right places).  This shows user that underlying engineering needs are exposed to the user and looks like poor design.  That is friction and complexity which is basically cyanide to adoption and growth of new contributions -- which means word of mouth and easy to understand for newbies is critical; every line of code counts, every verb / Noun matters.  Anything than can be removed from UX should be removed, with docs for power users that want to dig.
- saving and loading a graph without having to remember how `prepare!` is needed or what algorithm to use, just
```julia
lfg = load("desktop/gottiredyesterday.dfg")
# or
fg = NavAbilityDFG(identifier; api, auth)
lfg = merge!(LocalDFG(), fg)  # user already needs to dive into docs to learn merge! vs copy

solveGraph!(lfg; options=default) # maybe Monte Carlo via stateLabels

# save to come back to later
save("tmp.dfg", lfg)
sync!(fg, lfg)
```
- easily extensible with new factors (in `Main` context without having to recompile inside the library like all the competitors out there) -- we cannot ask someone that is trying to build their own factor to also deal with hooking into a bunch of obscure overloads.  The existing design requires only the residual and sampling functions via `CalcFactor`, that's it.  So adding anything that might require a user to do more than this for their custom factor will prevent UX'ers from becoming DX'ers.
```julia
# start like any other
fg = localdfg()
addFactor!(observation)

# oops i need a new factor for this other observation type
import IncrementalInference: CalcFactor, sample # from a docs tutorial
struct MyFactor <: AbstractRelative ...
(CalcFactor{<:MyFactor})(a,b,c) = residual(a,b,c)
sample(f::CalcFactor{<:MyFactor}) = rand(f.fancy)

# sceptical DX, how is this possible without recompile?  thanks Julia JIT
addFactor!(..., MyFactor(weird_data))

# DX utopia -- but not if there is even one friction call like `prepare!` or `init!`
solveGraph!(fg)
```

IncrementalInference/test/testBasicParametric.jl
Affie: `prepare` is an internal function that is **not exported** and **not public**, so not in user space.
The tests are busy testing internal functions and not user functions. 

> especially because it is solver specific, that makes it even worse

This is the exact reason implicit "prepare" needs to be taken out of `addVariable!` and `addFactor!` graph operations (violates the single responsibility principle). It does not make sense to need solver options set up to know how to add a node to the factor graph. The hack of having `:default` and `:parametric` select solvers is way worse IMHO and not aligned with other packages in the julia ecosystem.

A dedicated prepare phase remains structurally necessary to isolate graph definition from the allocation of solver-specific caches and memory buffers."

This same reasoning is for all "again", ie. test are busy testing internal functions, therefore, test has to use the modular building blocks separately. 

Perhaps we should explicitly split the tests in 2 categories: white(test/internal) and black box? This will make it explicit to users that are looking at the tests for examples.

IncrementalInference/src/services/SolverAPI.jl
Affie: All SDKs languages do not support `@async`.

IncrementalInference/src/services/FGOSUtils.jl
Affie: This is not related to this PR, it looks like the old hacky code that AMP v0.15 fixes.

IncrementalInference/src/parametric/services/ParametricManopt.jl
Affie: Just FIY, this was to fix an old intermittent Singular Exception bug that showed up again and prevented the test from passing.

Affie: The whole DebugTension part was quite handy to help find the issue.  I should have done it in separate PR though, but something about the changes made it show up agian.

IncrementalInference/src/services/GraphInit.jl
Affie: Yes, it is based on your requirement of having the "preperation" separate from the solving, so not JIT per node but per solve. This function makes the timing information you wanted possible.

IncrementalInference/test/testBasicParametric.jl
Affie: I don't have a problem with `initAll!` not in the UX, we can rename it to something better for the deprecation and not export the new function.

IncrementalInference/test/testSolveKey.jl
Affie: what would you have preferred? `setDefaultNodeData` (`NodeData`) is wrong in DFGv1, so would you have preffered `setDefaultState!`? (see the list of similar functions that used different verbs and nouns)

Also as I explained `NPBPSolver` is just the name I got form the documentation and I can rename to whatever you like by simple find and replace.

Affie: Just let me know what names you want and I'll rename.

IncrementalInference/test/testSpecialEuclidean2Mani.jl
Affie: What is the issue here?

IncrementalInference/src/services/GraphInit.jl
dehann: I'm assuming we can go for JIT per node/patch regardless of who needs it, i.e. if solve hits all nodes may end up doing cache/DL4-JIT...

IncrementalInference/src/services/SolverAPI.jl
dehann: On Julia supports solve, and we already decided that language specific syntax sugar is allowable deviations from the common signatures?  The 98% the same but don't limit Julia to useless python.

IncrementalInference/test/testBasicParametric.jl
dehann: > [SolverOptions replaces SolverParams]

Of course we need this PR and will merge.  This PR is just opening the next can of worms in the design towards IIF v1 (more than 3 months horizon).

> to isolate graph definition from the allocation of solver-specific caches and memory buffers.

A "Factor Graph" is a data-and-compute" object.  Not only a data object.  This is precisely the DataLevel 3 vs 4 issue that is vital to get right.  

> A dedicated prepare phase remains structurally necessary 

Maybe but that is contra-dispatch design (vision state comment), and regardless if the tech can or cannot handle it -- that is what the UX requires.  A solution must be found.  Remember, this is IIF (not DFG).  Solver cache will be stored in the graph somehow -- let's use this PR as consolidate work in progress.

By contra-dispatch I mean if you have an object `1//2` and need the float value `~0.5`, then auto promote or convert is available through dispatch.  The 'structurally necessary data preparation' happens implicitly through the functions available to the compiler.  In contrast ObjOriented design says, you need to run all these steps until the object is in the right state -- which is the opposite of the intention of separating data from logic.  Saying I have object `A` and must manually ensure it is prepped to be `A+` before the next thing is the cornerstone of why dispatch exists -- i.e. `promote/convert/constructor(::A+, a:;A)`. 

I'm wary about "structurally necessary".  Also acknowledging much uglier code I wrote, going back the first 3 generations of `CommonConvWrapper / GenericFunctionNodeData{ComputeCache} / etc`.

> prepare is an internal function that is not exported and not public, so not in user space.
The tests are busy testing internal functions and not user functions.
> This same reasoning is for all "again", 

Maybe okay in a few tests, but something is wrong if it's being spread across many tests -- that sounds like it will become a UX requirement. 

> Perhaps we should explicitly split the tests in 2 categories: white(test/internal) and black box? This will make it explicit to users that are looking at the tests for examples.

let's come back to whether we really want that.  I'm still in the no camp.  DFG/IIF/CJL is not a one size fits all system.  Word of mouth should be "easy install and stuff works / results are state of the art + permissive open-science/core".  Anything more and viral sharing will not occur.  If User++ wants to build their own solver variants, great and it is easy for them to do -- that is attractive, but...  There should be no penalty on users who don't want to build their own solver against this code, hence zero additional friction should exist for all users just because some users want a one-size-fits all.

xref: I've been having some success in AMP with `const HomotopyDensity = Union{<:HomotopyDensityDFG, HomotopyDensityLive}` -- still early days to see if that is a workable pattern for solving DataLevel 3.5 desires.

----

## In-situ Notes Capture

Strictly speaking, when a user runs `solveGraph!`, all they want is the best possible answer that we can provide.  It's bad to say to all users, "hey, our solution is not good enough, here is an easy dial to change between other sucky solutions..." if that happens then there will be no users.  The origin of `:default` aims towards common effort on one solver.  `:parametric` was a hack to try work quickly on a separate solver, now the cart is pulling the horse.  Yes, there are different solvers, but basically what should happen is if a user tries any other`SolverOption` (other than defaults) then they should get a worse answer.

If user tries `SolverOption(GTSAMEquivalent())`, they should see after analysis that the answer is worse.  Similarly, if they try straight Riemannian Levenberg-Marquardt -- again the results should be worse.  We are engineering a default solver which is better -- if we are not focused on building one better solver, then there is no point.  Reminder, the road towards hybrid-(non)parametric / continuation-deconv solution has a few steps:
1. step one, get a common data format whereby both old school parametric and nonparametric methods can come closer together (aka DFG v1 w `HomotopyDensity`)
1. further blur the lines towards a solver which only grows extra modes (or nonparametric behavior) where necessary -- simultaneously collapse obviously unimodal states, BUT **without becoming a [greedy/locally optimal algorithm](https://en.wikipedia.org/wiki/Greedy_algorithm)** -- this is a hard problem, good news is we already have a unusually strong solution to this hard problem.  I don't want users hunting and pecking around like it's a bazaar, getting distracted that the default solver is state-of-the art.  Of course users should be able to change solvers, but it's not our job to just build a database for other solvers -- the project will fail if we do that.  That exists and is called G2O (missing Ceres when i last checked).
1. Clique by clique solves using `solveClique!` in a highly scalable compute environment.
1. Switch to full clique dimension deconvolution solves rather than convolution solving strategies.
1. Incrementally improve on clique-level-deconvolution strategy, which I expect will be continuation theory based.

We just finshed 1.) and this and other PRs happening are the start of 2.)

IncrementalInference/test/testBasicParametric.jl
dehann: dream here is just init on the tree.  I believe it is possible to only do numerics once transiting the Bayes tree.  One day we might even get to numerics only on down passes of the tree (connecting with the old X-stroke conversation we had).

IncrementalInference/test/testSpecialEuclidean2Mani.jl
dehann: not sure what triggered here -- this comment can resolve

IncrementalInference/test/testSolveKey.jl
dehann: thanks, but the point is that ultimately there shouldn't be different `prepareState!`s -- that complexity should move upstream into a common interface functions of `HomotopyDensity`, and maybe a solver specific `prepareCache` -- at node/patch level atomic.

If two different solvers are producing apples and oranges states, something is wrong in the design. I know this is work in progress to consolidate -- im talking vision state here.

PR #1956, opened 2026-08-18
Merge Parametric tree solve from release to develop

Affie:
I'm not sure what the state of develop currency is, but i think the failing tests are unrelated to this merge.
Current results:
>Some tests did not pass: 1152 passed, 9 failed, 27 errored, 41 broken.

I fixed tree related functions that worked in release branch that caused the tests to hang indefinitely.


Affie:
I'm going to merge as is, will just fix anything i broke in the merge separately as it doesn't look everything was working before the merge.

